PR #6, opened 2023-09-19
Add https and wss support

### LINKED ISSUE #4: Remote server https and wss not supported
I am running comfy ui on a cloud server and it exposes only https and wss.

This can be easily resolved by editing client.py and network.py. A checkbox or dropdown list to choose between secure connection would be the simplest.

https://github.com/Acly/krita-ai-diffusion/blob/8e14748df6d2c0ee0255882b715f3dfa2e0ad7be/ai_diffusion/client.py#L110
https://github.com/Acly/krita-ai-diffusion/blob/8e14748df6d2c0ee0255882b715f3dfa2e0ad7be/ai_diffusion/network.py#L65

gergelyszaz:
Fixes #4

### DIFF
--- a/ai_diffusion/client.py
+++ b/ai_diffusion/client.py
@@ -83,7 +83,7 @@ class DeviceInfo(NamedTuple):
 class Client:
     """HTTP/WebSocket client which sends requests to and listens to messages from a ComfyUI server."""
 
-    default_url = "127.0.0.1:8188"
+    default_url = "http://127.0.0.1:8188"
 
     _requests = RequestManager()
     _websocket: websockets.WebSocketClientProtocol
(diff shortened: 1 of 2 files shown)


gergelyszaz:
Only tested the https and wss connection. Should still work if not the full url is provided, only the hostname.

gergelyszaz:
Might as well trim the trailing slash in an url.
https://awesome-site.com/ is still problematic.

gergelyszaz:
Something is not right yet... the connection is created but I don't get the image.

Acly:
Hm I also use extra_model_paths.yaml and had duplicates before without issues I think.

Anyway, PR looks good, thank you!

PR #9, opened 2023-09-19
Add new style and delete style buttons

### LINKED ISSUE #7: Feature request: Add option to create a new style preset from Krita
Currently I am only able to add a new one by copying an existing style. 
A simple add button would suffice for me. 
Optionally the file name should also be renamed whenever the style name changes.

gergelyszaz:
Closes #7 

Adds create style and delete style buttons.
image

The create style opens an input dialog where the new style name can be given.
image
After the stlye is created, it becomes the currently selected style.

Delete style removes the currently loaded style. There is NO confirmation dialog currently (I don't need it, but could be added).

In both cases the style list is refreshed.

### DIFF
--- a/ai_diffusion/style.py
+++ b/ai_diffusion/style.py
@@ -1,5 +1,6 @@
 from enum import Enum
 import json
+import os
 from pathlib import Path
 from PyQt5.QtCore import QObject, pyqtSignal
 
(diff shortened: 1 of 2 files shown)


gergelyszaz:
Maybe the buttons don't need to be member fields and could be just local variables.

Acly:
> Maybe the buttons don't need to be member fields and could be just local variables.

Technically they don't have to be, but it doesn't hurt imo.

I like the buttons. I'd prefer a solution without popup dialog (eg. have an automatic name like "New Style" and use Name field to adjust), but it's more work because of file renaming, and not really important.

What happens if you choose a name that clashes with already existing file? They overwrite each other?
What happens for names that don't make valid filenames?

ai_diffusion/style.py
```diff
@@ -193,14 +194,27 @@ def __init__(self, folder: Path = default_folder):
...
+        new_style = Style(self.folder / f"{name}.json")
+        new_style.name = name
+        self._list.append(new_style)
+        new_style.save()
+        self.changed.emit()
+        return new_style
+    
+    def delete(self, style: Style):
+        self._list.remove(style)
+        os.remove(style.filepath)
```
Acly, line 208: or `style.filepath.unlink()`

gergelyszaz:
Have not tested that case, but I think it overrides it. I'll add an error popup instead.

gergelyszaz:
Existing name: is overwritten
Invalid name: nothing happens

I start to like your suggestion:
- no popup
- default name: style_{i}.json
- increment i until no file with that name can be found

Acly:
Looks good, thanks. I had to fix the default cfg value to not be float (QSlider doesn't like it), but I'll just put that on top.

PR #106, opened 2023-11-22
Feature: Controlnet QR Code Monster

jellydreams:
Added the option [QR Code Monster](https://huggingface.co/monster-labs/control_v1p_sd15_qrcode_monster) to the controlnet list (without a preprocessor)
for **model**: [control_v1p_sd15_qrcode_monster](https://huggingface.co/monster-labs/control_v1p_sd15_qrcode_monster/blob/main/control_v1p_sd15_qrcode_monster.safetensors) 

krita_qrcodemonster


Acly:
Thanks for the PR, yes we can add QR Code Monster.

Some remarks/questions:
1. I would prefer a name which describes what the CN does. QRCodeMonster might be known among those who follow SD, but the plugin should be accessible for people who have never heard of it. I suggest "Pattern" (or maybe "Stencil"). Could use a QR code as icon instead, best of both worlds.
2. The generate (preprocessor) button should not be visible since it doesn't do anything. Same as "image" control mode https://github.com/Acly/krita-ai-diffusion/blob/main/ai_diffusion/ui/widget.py#L241
3. There appears to be a "v2" version of the model which is supposedly better, have you tried it? Just a question, haven't compared it myself.

jellydreams:
1. I vote for 'Stencil'; it seems to be a good name that describes it for everyone 👍

2. Thanks for linking the line; it was easy to remove the button
image

3. I swapped the file `control_v1p_sd15_qrcode_monster.safetensors` with `control_v1p_sd15_qrcode_monster_v2.safetensors`, and it still works (and doesn't work if the file is absent). I've tried them too little to tell the difference.

ai_diffusion/resources.py
```diff
@@ -256,6 +256,14 @@ class ModelResource(NamedTuple):
         "control_lora_rank128_v11p_sd15_seg_fp16.safetensors",
         "https://huggingface.co/comfyanonymous/ControlNet-v1-1_fp16_safetensors/resolve/main/control_lora_rank128_v11p_sd15_seg_fp16.safetensors",
     ),
+    ModelResource(
+        "Controlnet Stencil",
+        ResourceKind.controlnet,
+        SDVersion.sd15,
+        Path("models/controlnet"),
+        "control_v1p_sd15_qrcode_monster.safetensors",
+        "https://huggingface.co/monster-labs/control_v1p_sd15_qrcode_monster/blob/main/control_v1p_sd15_qrcode_monster.safetensors",
```
Acly, line 265: I believe this links to a web page - it should be a direct download link.

Acly:
Any idea why the icon is always black in the UI?

jellydreams:
I can replace with png, but is there a solution with svg ? It would provide more flexibility.

For svg icons:
- Is there a section in the code that sets the style of icons based on the theme?
- Is there a list of resources to which file names must be added ?
- If not, are there specific steps or checks to perform when exporting svg files with Inkscape?

Acly:
To be clear, the .svg you added is kind of working, and no code is needed: it loads `f"control-{mode.name}-{theme}.svg"`, because the filename matches it works.
Just for some reason while you can see the butterfly shape, it doesn't seem to have the correct color. Also in your screenshot. I tried opening in Inkscape, and it displayed the color correctly there, but I wasn't able to edit it for some reason.

I made the other icons in Inkscape too, and didn't do anything special, just set the fill or line color on the shapes.

jellydreams:
Thanks for the clue ! It seems that there were some XML lines in svg overwriting the color, it should work now:
image
image

Issue #69, opened 2023-11-19
Start and End of Influence (Controlnet)


Luzido:
First of all: Great Plugin!! awesome.

I think a limiting factor right now is that we cannot adjust the start and end of the controlnet influence. I had really awesome results setting lineart cnet to end at 0.3 or something.
I wish that would be possible!

Acly:
I've looked into it a bit, but I'm not convinced it can fit in directly. There is a lot of overlap with ControlNet strength, and I don't want to clutter the UI with 3 configurable values. 

Not saying a case can't be made for it, but it should be motivated by
* use cases where you need start/end and one strength value isn't enough
* an intuition about those values: start/end is very technical and you need to know about sampling process, but it could be presented as "how much it affects composition" and "how much it affects details"
* some examples on what you can achieve with it (advantages compared to plain CN with strength tweaking)

It would also be interesting if quality can be improved by adjusting all of start/end/strength under the hood of one user-exposed value.

Acly:
ControlNet end parameter is now supported in 1.8.0

PR #107, opened 2023-11-23
ControlNet ending step support (and scribble update)

Danamir:
## Concept

I propose to add a setting field to each control layer to select the ending step ratio. This setting can be combined with strength to achieve better and faster results. Generally speaking, we want the ControlNet to be applied to the first steps to alter the structure of the image, leaving the later steps alone to do the detailing. It even improves performance because in ComfyUI if you set for example the end_step to 0.5, the ControlNet performance impact only occurs on half the steps. 

I also replaced the FakeScribble preprocessor by PiDiNet + Scribble preprocessors, as the results are much more viable as a sketch input to be quickly updated with a simple brush. The SDXL sketch ControlNet can be used for "Scribble" **and** "Lineart" control modes with great results.

(NB : sorry if the two PR are mixed, I was using the scribble control mode as a testing point for ControlNet ending step setting as I am more familiar with it since I set it as default in the SdPaint development. I can rollback this part if needed.)

## Interface

Here is the setting in place : 

image

I also added a configuration setting to hide this field :  

image

I left it to true by default, but of course it could be set to false to keep the interface unchanged.

## Results

Here are some outputs comparison with ControlNet sketch with the original prompt "_A bearded man portrait, comics illustration_", updated prompt "_A bearded viking man portrait, comics illustration_" and seed 42. This method can be applied to all the control methods, personally I generally don't go above 0.7 end step to avoid rendering artifacts.

 - Original picture and detected PiDiNet scribble : 

image 
image

 - Updated picture with control scribble at 100% Strength and 1.0 end step ratio : 
 
You can see that the ControlNet is too strong at the end and produces graphical artifacts on the lines (those are somewhat too sharp).
 
image

 - Updated picture with control scribble at 60% Strength and 1.0 end step ratio : 

As expected the sketch is less followed, but the line art is still too sharp compared to the original picture.

image

- **Lastly with the updated code**, 100% Strength and 0.6 end step ratio : 

The sketch is almost perfectly followed, but the final rendering steps produce fewer graphical artifacts.

image

## Performances

Lowering strength does not affect the performance, but the ending step provides a boost to later steps as those are run without ControlNet interference.

 - 100% strength & 1.0 end :  
```
Requested to load SDXL
100%|██████████| 20/20 [00:14<00:00,  1.39it/s]
Prompt executed in 18.01 seconds
```

 - 60% strength & 1.0 end : 
```
Requested to load SDXL
100%|██████████| 20/20 [00:14<00:00,  1.39it/s]
Prompt executed in 17.79 seconds
```

- 100% strength & 0.6 end : 
```
Requested to load SDXL
100%|██████████| 20/20 [00:14<00:00,  1.70it/s]
Prompt executed in 15.18 seconds
```

## Todo
 - [x] Hide ending setting for `ControlMode.image` controls
 - [x] Set interface setting to `false` by default
 - [x] Update control `scribble` test reference picture

### DIFF
--- a/ai_diffusion/ui/widget.py
+++ b/ai_diffusion/ui/widget.py
@@ -139,7 +139,7 @@ class ControlWidget(QWidget):
         self.end_spin.setValue(1.0)
         self.end_spin.setSuffix("")
         self.end_spin.setSingleStep(0.1)
-        self.end_spin.setToolTip("Control end step ratio")
+        self.end_spin.setToolTip("Control ending step ratio")
         self.end_spin.valueChanged.connect(self._notify)
         self.end_spin.setVisible(settings.show_control_end)
 
--- a/ai_diffusion/ui/widget.py
+++ b/ai_diffusion/ui/widget.py
@@ -141,6 +141,7 @@ class ControlWidget(QWidget):
         self.end_spin.setSingleStep(0.1)
         self.end_spin.setToolTip("Control end step ratio")
         self.end_spin.valueChanged.connect(self._notify)
+        self.end_spin.setVisible(settings.show_control_end)
 
         self.error_text = QLabel(self)
         self.error_text.setText("ControlNet not installed")
@@ -255,7 +256,7 @@ class ControlWidget(QWidget):
         self.add_pose_button.setEnabled(self._is_vector_layer())
         self.strength_spin.setVisible(is_installed)
         self.strength_spin.setEnabled(self._is_first_image_mode())
-        self.end_spin.setVisible(is_installed)
+        self.end_spin.setVisible(is_installed and settings.show_control_end)
         self.end_spin.setEnabled(self._is_first_image_mode())
         self.error_text.setVisible(not is_installed)
         return is_installed
@@ -333,6 +334,12 @@ class ControlListWidget(QWidget):
     def _remove_widget(self, control: ControlWidget):
         self._controls.remove(control)
         control.deleteLater()
+
+    def update_control_field(self, name, function):
+        for control in self._controls:
+            setting = getattr(control, name, None)
+            if setting is not None:
+                function(setting)
 
 
 class ControlLayerButton(QToolButton):
@@ -720,6 +727,9 @@ class GenerationWidget(QWidget):
         elif key == "show_negative_prompt":
             self.negative_textbox.clear()
             self.negative_textbox.setVisible(value)
+        elif key == "show_control_end":
+            self.control_list.update_control_field("end_spin", lambda x: x.setVisible(value))
+            self.control_list.update_control_field("end_spin", lambda x: x.setValue(1.0))
 
     def show_results(self, job: Job):
         if job.kind is JobKind.diffusion:
--- a/ai_diffusion/ui/widget.py
+++ b/ai_diffusion/ui/widget.py
@@ -256,7 +256,7 @@ class ControlWidget(QWidget):
         self.add_pose_button.setEnabled(self._is_vector_layer())
         self.strength_spin.setVisible(is_installed)
         self.strength_spin.setEnabled(self._is_first_image_mode())
-        self.end_spin.setVisible(is_installed and settings.show_control_end)
+        self.end_spin.setVisible(is_installed and settings.show_control_end and mode is not ControlMode.image)
         self.end_spin.setEnabled(self._is_first_image_mode())
         self.error_text.setVisible(not is_installed)
         return is_installed
--- a/ai_diffusion/ui/widget.py
+++ b/ai_diffusion/ui/widget.py
@@ -131,7 +131,16 @@ class ControlWidget(QWidget):
         self.strength_spin.setValue(100)
         self.strength_spin.setSuffix("%")
         self.strength_spin.setSingleStep(10)
+        self.strength_spin.setToolTip("Control strength")
         self.strength_spin.valueChanged.connect(self._notify)
+        
+        self.end_spin = QDoubleSpinBox(self)
+        self.end_spin.setRange(0.0, 1.0)
+        self.end_spin.setValue(1.0)
+        self.end_spin.setSuffix("")
+        self.end_spin.setSingleStep(0.1)
+        self.end_spin.setToolTip("Control end step ratio")
+        self.end_spin.valueChanged.connect(self._notify)
 
         self.error_text = QLabel(self)
         self.error_text.setText("ControlNet not installed")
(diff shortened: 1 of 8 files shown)


ai_diffusion/ui/widget.py
```diff
@@ -243,6 +256,8 @@ def _check_is_installed(self):
         self.add_pose_button.setEnabled(self._is_vector_layer())
         self.strength_spin.setVisible(is_installed)
         self.strength_spin.setEnabled(self._is_first_image_mode())
+        self.end_spin.setVisible(is_installed and settings.show_control_end)
+        self.end_spin.setEnabled(self._is_first_image_mode())
```
Acly, line 260: This isn't needed, or rather I think we shouldn't show the end spin box at all for "image" control layer.
That is, add `and mode is not ControlMode.image` to the `setVisible` condition.

Note: ControlMode.image is how IP-adapter is exposed in the UI, it's not technically a ControlNet and applied differently.

Acly:
It would be nice if you could start up ComfyUI server and run `pytest tests` in the repo dir. Or `pytest tests\test_workflow.py::test_create_control_image` in particular. It will probably fail because scribble preprocessor generates a different image now, and replacing the reference image should fix it.

If you have trouble running the test let me know, they might not be super robust.

ai_diffusion/ui/widget.py
Danamir: > Note: ControlMode.image is how IP-adapter is exposed in the UI, it's not technically a ControlNet and applied differently.

Oh, I see. I was not sure what was behind `ControlMode.image`. I'll do the modification.

On a side note, a name change for this particular control could clarify it's usage for anyone wondering. 😅

Danamir:
Do you have a `requirements.txt` file for the project and/or the tests part ? I'm trying to install the missing libraries manually but I may miss some.

Acly:
Thank you for the PR btw, great comparison. There is a feature request #69 where I explained my opinion on additional UI for the end ratio. I'm generally quite convinced that using the end ratio can improve results. I am just not sure yet about how to expose it: adding another number box is one too many I feel, and its description isn't intuitive unless you have technical knowledge about the sampling process.

So I would suggest for now, to keep the setting on "Don't show" by default. I'm okay with merging this PR as a way to tweak it for users who know how it works. But I see it as a bit of an intermediate step.

For future work, the best solution IMO would be if there was just one number, but it modifies both end-ratio and strength to get best results. Not necessarily in a linear way.
str

Does that make sense? Default "UI strength" should probably be around 80%.

Testing that and finding good values for all the different control net is certainly a lot of work though. But I think it could give  people better results by default. And I'm not actually sure even power users need the possibility to tweak both independently.

Acly:
> Do you have a `requirements.txt` file for the project and/or the tests part ? I'm trying to install the missing libraries manually but I may miss some.

No, but it's a good idea. It shouldn't actually be many, since Krita's embedded python is very bare bones. I'll try and check which ones are actually required.

Danamir:
Only the test scribble seems to be failing : 

```
Launching pytest with arguments test_workflow.py::test_create_control_image --no-header --no-summary -q in F:\SD\krita-ai-diffusion\tests

============================= test session starts =============================
collecting ... collected 8 items

test_workflow.py::test_create_control_image[ControlMode.scribble] 
test_workflow.py::test_create_control_image[ControlMode.line_art] 
test_workflow.py::test_create_control_image[ControlMode.soft_edge] 
test_workflow.py::test_create_control_image[ControlMode.canny_edge] 
test_workflow.py::test_create_control_image[ControlMode.depth] 
test_workflow.py::test_create_control_image[ControlMode.normal] 
test_workflow.py::test_create_control_image[ControlMode.pose] 
test_workflow.py::test_create_control_image[ControlMode.segmentation] 

======================== 1 failed, 7 passed in 13.98s =========================
FAILED [ 12%]
test_workflow.py:362 (test_create_control_image[ControlMode.scribble])
0.31692815 != 0.001

Expected :0.001
Actual   :0.31692815
```

Danamir:
> For future work, the best solution IMO would be if there was just one number, but it modifies both end-ratio and strength to get best results. Not necessarily in a linear way.
> 
> Does that make sense? Default "UI strength" should probably be around 80%.
> 
> Testing that and finding good values for all the different control net is certainly a lot of work though. But I think it could give people better results by default. And I'm not actually sure even power users need the possibility to tweak both independently.

That's actually a great idea. I was tempted to do it in this PR but I didn't want to add too many things at once.

By experience in SdPaint I fine-tuned the strength/ending step presets further on my own install than in the default presets to quickly switch between ones that I can kinda think of "light control" "heavy control" "locked" etc. But I even got as far as using 150% strength and 0.2 ending step in some fringe cases. Sadly it's heavily dependent on the ControlNet model and SD1.5 or SDXL version used. The one I use the most for SD 1.5 scribble is 80% str & 0.5 step as a good middle ground. 

If we want to simplify the user experience we could replace the actual setting strength by a fev (4-5ish) presets. This could be done per control and per model generation (sd15 or sdxl). But it would be harder to maintain for any future control models. A more realistic approach would be to have those few presets and let the user explore which one is good at which setting for which control/model. 

> So I would suggest for now, to keep the setting on "Don't show" by default. I'm okay with merging this PR as a way to tweak it for users who know how it works. But I see it as a bit of an intermediate step.

Exactly. For this PR I think that simply leaving the interface setting to `false` by default can suffice. The presets can be done in a future PR. I'll do the modification later today.

Acly:
> Oh I see, the scribble is failing because of my modifications on the preprocessors. I can update the image with the new one if you want.

Yes, please replace the image

Danamir:
Reference image updated, tests ok : 

```
============================= test session starts =============================
collecting ... collected 8 items

test_workflow.py::test_create_control_image[ControlMode.scribble] 
test_workflow.py::test_create_control_image[ControlMode.line_art] 
test_workflow.py::test_create_control_image[ControlMode.soft_edge] 
test_workflow.py::test_create_control_image[ControlMode.canny_edge] 
test_workflow.py::test_create_control_image[ControlMode.depth] 
test_workflow.py::test_create_control_image[ControlMode.normal] 
test_workflow.py::test_create_control_image[ControlMode.pose] 
test_workflow.py::test_create_control_image[ControlMode.segmentation] 

============================= 8 passed in 15.00s ==============================
```

```
============================= test session starts =============================
collecting ... collected 5 items

test_workflow.py::test_control_scribble[generate] 
test_workflow.py::test_control_scribble[inpaint] 
test_workflow.py::test_control_scribble[refine] 
test_workflow.py::test_control_scribble[refine_region] 
test_workflow.py::test_control_scribble[inpaint_upscale] 

============================= 5 passed in 49.42s ==============================
```

Danamir:
Done.

Do you see other relevant modifications necessary ?

Acly:
Looks good! Will merge soon.

PR #112, opened 2023-11-24
Performance upgrade by using KSamplerAdvanced

Danamir:
Replacing `KSampler` by `KSamplerAdvanced` can significantly increase the performance, by using the `add_noise` and `start_at_step` parameters instead of the `denoise` one. The generated images are very similar. 

I'll document the changes in later comments.

 - [x] Use advanced sampler setting
   - Set as new default and remove the setting ?
 - [x] Implement KSamplerAdvanced denoising
 - [x] Skip sampler_steps_upscaling parameter
 - [x] Test rendering
   - [x] Generate
     - [x] LoRA
   - [x] Inpaint
   - [x] Controls
     - [x] IPAdapter
     - [x] OpenPose
     - [x] Sketch
     - [x] Depth
   - [x] ✖ Upscale 
   - [x] ❌ Live
     - [x] Disable for LCM/Live

### DIFF
--- a/ai_diffusion/comfyworkflow.py
+++ b/ai_diffusion/comfyworkflow.py
@@ -103,16 +103,10 @@ class ComfyWorkflow:
         sampler="dpmpp_2m_sde_gpu",
         scheduler="normal",
         steps=20,
+        start_at_step=0,
         cfg=7.0,
-        denoise=1.0,
         seed=-1,
-        min_steps=None,
     ):
-        start_at_step = round(steps*(1-denoise))
-        if min_steps and steps - start_at_step < min_steps:
-            start_at_step = math.floor(steps * 1/denoise - steps)
-            steps = start_at_step + min_steps
-
         self.sample_count += steps - start_at_step
 
         return self.add(
--- a/ai_diffusion/comfyworkflow.py
+++ b/ai_diffusion/comfyworkflow.py
@@ -94,6 +94,39 @@ class ComfyWorkflow:
             denoise=denoise,
         )
 
+    def ksampler_advanced(
+        self,
+        model: Output,
+        positive: Output,
+        negative: Output,
+        latent_image: Output,
+        sampler="dpmpp_2m_sde_gpu",
+        scheduler="normal",
+        steps=20,
+        cfg=7.0,
+        denoise=1.0,
+        seed=-1,
+    ):
+        self.sample_count += steps
+
+        return self.add(
+            "KSamplerAdvanced",
+            1,
+            noise_seed=random.getrandbits(64) if seed == -1 else seed,
+            sampler_name=sampler,
+            scheduler=scheduler,
+            model=model,
+            positive=positive,
+            negative=negative,
+            latent_image=latent_image,
+            steps=steps,
+            start_at_step=round(steps*(1-denoise)),
+            end_at_step=steps,
+            cfg=cfg,
+            add_noise='enable',
+            return_with_leftover_noise='disable',
+        )
+
     def model_sampling_discrete(self, model: Output, sampling: str):
         return self.add("ModelSamplingDiscrete", 1, model=model, sampling=sampling, zsnr=False)
 
--- a/ai_diffusion/comfyworkflow.py
+++ b/ai_diffusion/comfyworkflow.py
@@ -108,12 +108,12 @@ class ComfyWorkflow:
         seed=-1,
         min_steps=None,
     ):
-        self.sample_count += steps
-
         start_at_step = round(steps*(1-denoise))
         if min_steps and steps - start_at_step < min_steps:
             start_at_step = math.floor(steps * 1/denoise - steps)
             steps = start_at_step + min_steps
+
+        self.sample_count += steps - start_at_step
 
         return self.add(
             "KSamplerAdvanced",
--- a/ai_diffusion/comfyworkflow.py
+++ b/ai_diffusion/comfyworkflow.py
@@ -112,7 +112,7 @@ class ComfyWorkflow:
 
         start_at_step = round(steps*(1-denoise))
         if min_steps and steps - start_at_step < min_steps:
-            start_at_step = round(steps * 1/denoise - steps)
+            start_at_step = math.floor(steps * 1/denoise - steps)
             steps = start_at_step + min_steps
 
         return self.add(
--- a/ai_diffusion/comfyworkflow.py
+++ b/ai_diffusion/comfyworkflow.py
@@ -106,8 +106,14 @@ class ComfyWorkflow:
         cfg=7.0,
         denoise=1.0,
         seed=-1,
+        min_steps=None,
     ):
         self.sample_count += steps
+
+        start_at_step = round(steps*(1-denoise))
+        if min_steps and steps - start_at_step < min_steps:
+            start_at_step = round(steps * 1/denoise - steps)
+            steps = start_at_step + min_steps
 
         return self.add(
             "KSamplerAdvanced",
@@ -120,7 +126,7 @@ class ComfyWorkflow:
             negative=negative,
             latent_image=latent_image,
             steps=steps,
-            start_at_step=round(steps*(1-denoise)),
+            start_at_step=start_at_step,
             end_at_step=steps,
             cfg=cfg,
             add_noise='enable',
--- a/ai_diffusion/style.py
+++ b/ai_diffusion/style.py
@@ -121,12 +121,6 @@ class StyleSettings:
         "Higher values can produce more refined results but take longer",
     )
 
-    sampler_steps_upscaling = Setting(
-        "Sampler Steps (Upscaling)",
-        15,
-        "Additional sampling steps to run when automatically upscaling images",
-    )
-
     cfg_scale = Setting(
         "Guidance Strength (CFG Scale)",
         7.0,
@@ -156,7 +150,6 @@ class Style:
     vae: str = StyleSettings.vae.default
     sampler: str = StyleSettings.sampler.default
     sampler_steps: int = StyleSettings.sampler_steps.default
-    sampler_steps_upscaling: int = StyleSettings.sampler_steps_upscaling.default
     cfg_scale: float = StyleSettings.cfg_scale.default
     live_sampler: str = StyleSettings.live_sampler.default
     live_sampler_steps: int = StyleSettings.live_sampler_steps.default
@@ -210,7 +203,7 @@ class Style:
         if is_live:
             return SamplerConfig(self.live_sampler, self.live_sampler_steps, self.live_cfg_scale)
         if is_upscaling:
-            return SamplerConfig(self.sampler, self.sampler_steps_upscaling, self.cfg_scale)
+            return SamplerConfig(self.sampler, self.sampler_steps, self.cfg_scale)
         return SamplerConfig(self.sampler, self.sampler_steps, self.cfg_scale)
 
 
--- a/ai_diffusion/style.py
+++ b/ai_diffusion/style.py
@@ -199,11 +199,9 @@ class Style:
     def filename(self):
         return self.filepath.name
 
-    def get_sampler_config(self, is_upscaling=False, is_live=False):
+    def get_sampler_config(self, is_live=False):
         if is_live:
             return SamplerConfig(self.live_sampler, self.live_sampler_steps, self.live_cfg_scale)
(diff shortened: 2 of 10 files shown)


Danamir:
The default settings use the old method, so the tests are ok : 

```
============================= test session starts =============================
collecting ... collected 3 items

test_workflow.py::test_generate[extent0] 
test_workflow.py::test_generate[extent1] 
test_workflow.py::test_generate[extent2] 

============================= 3 passed in 23.05s ==============================
```

Danamir:
## Comparisons

Source image : 
image

 - 50% dog with old KSampler method : 
```
100%|██████████| 20/20 [00:09<00:00,  2.17it/s]
Prompt executed in 13.44 seconds
```
image

 - 50% dog with new KSamplerAdvanced method:
```
100%|██████████| 10/10 [00:04<00:00,  2.16it/s]
Prompt executed in 7.74 seconds
```
image

 - Other results :  
image

Danamir:
Using inpaint on the face only, with a control layer on image (thus IPAdapter) mode at 50% strength, and 60% global denoise. The results are also similar.

 - Old method:
```
100%|██████████| 20/20 [00:10<00:00,  1.94it/s]
Prompt executed in 23.38 seconds
```
image

 - New method:
 ```
100%|██████████| 12/12 [00:10<00:00,  1.94it/s]
Prompt executed in 9.43 seconds
```
image

**Comment by Danamir** [comment]
## Notes

The `KSamplerAdvanced` also offers the possibility of using `add_noise` and `return_with_leftover_noise` to chain the samplers to do a refine pass.

This could for example be used in SDXL to use a `ClipTextEncodeSDXLRefiner` with aesthetic score parameter. Or one thing that I do in most of my workflows : a first DPM++ SDE pass for 50-60% of the steps as I prefer its structure and prompt comprehension, and the rest of the steps with a faster sampler to handle the later rendering. 

But I admit this is a very advanced use case, and used only for the base sampling, not the inpaint. So it may have more its place on a personal fork of mine, leaving the main plugin free of clutter.

**Comment by Danamir** [comment]
I did some testing on live/LCM mode. It kinda works, bu the low steps (ie. 6) does not allow for enough nuance on the denoising setting. So if you go from 50% to 60% the step goes from 3 to 4 and there is a major difference in rendering with `KSamplerAdvanced`. Where with `KSampler` if you go from 6 steps with 50% noise to 60% noise the difference is much more subtle.

I will fallback to the normal `KSampler` method for the live mode.

**Comment by Danamir** [comment]
This method does not affect `upscale_tiled`, because it uses the `UltimateSDUpscale`.

Theoretically it should work with `upscale_simple`, but this method does not seem to be reachable. (`UpscaleParams.use_diffusion` is affected by `use_refinement`, but if `use_refinement` is unchecked there is no sampling called if I'm not mistaken ?)

**Comment by Acly** [comment]
Very interesting, I was implementing at some point a version that scaled the steps according to denoising. Results were not quite as good for low denoise, so didn't persue. I did not use the advanced sampler, I remember reading somewhere that it boils down to the same thing (ie. 10 steps 50% denoise -> advanced 20 steps total, start at 10). But probably that was only concerning how much noise is applied, setting the start step seems to result in different schedule.

I'll give it a try, but looks promising. I don't think we want to support two different options. If there are issues in some cases because we end up with too few steps it's always possible to increase them slightly (but not to the full amount that would be used for 100% strength).

**Comment by Acly** [comment]
> Theoretically it should work with `upscale_simple`, but this method does not seem to be reachable. (`UpscaleParams.use_diffusion` is affected by `use_refinement`, but if `use_refinement` is unchecked there is no sampling called if I'm not mistaken ?)

Yes `upscale_simple` does not sample, it just runs an upscaling model.

**Comment by Danamir** [comment]
>  If there are issues in some cases because we end up with too few steps it's always possible to increase them slightly

I've been using this method in ComfyUI for some time now, and I think Auto1111 does something similar (it's not rare to see 1-2 steps at very low denoising) and the results are almost identical. At higher denoising/higher steps count the differences are more visible, but much lower than if a different seed is selected.

If we want to keep only the `KSamplerAdvanced`, I think the only thing we need is a `min_steps` parameter to be used in live mode. It would calculate the new starting step and total steps to keep the difference above this minimum.

I did some tests with extreme values : imagine a case where the steps to 6 and strength is set to 10%, thus to keep the minimum steps to 6 we would have to set the total steps to 66 and the starting step to 60. But the results are identical at 6/10% vs. 60/66, and 6/20% vs. 30/36, and 6/30% vs. 20/26.

I'll code this solution to give it a try.

**Comment by Danamir** [comment]
The live/LCM mode now uses 6 steps minimum (set by the style settings), and the results are exactly the same as with the simple `KSampler`.

With the old method, there are some strength settings that give the same results (ie. 95-90%, 85-80%, 75-70%). Same with the new method.

**Comment by Danamir** [comment]
Notes on DPM++ 2M SDE : this sampler is more prone to artifacts, especially when used at low steps and/or high CFG and with LoRAs (ie. < 18 steps, and > 6 CFG).

So when used with the default settings of 20 steps and 7.0 CFG, around 70-80% strength **and** with a LoRA some artifacts may occur because the `KSamplerAdvanced` method is also lowering the steps. At higher strength the problem is not present because of the steps number, neither at lower steps because of the influence of the source image.

I this very particular case, lowering the CFG to 6.0 solves the problem, or defaulting to DPM++ 2M of course.

I don't know if this needs a fix in the code, or if it is left to the user to know the particular limitations of this sampler.

**Review by Acly — COMMENTED** [review]
I did some testing and it worked really well! So let's remove the option and make advanced ksampler the default.

I would also like to suggest moving the start-at-step computation to `_sampler_params`, or its own function which is called from there. It's a good candidate for a test too, to check the corner cases.

File: ai_diffusion/comfyworkflow.py [hunk 1]
```diff
@@ -94,6 +94,45 @@ def ksampler(
...
+        latent_image: Output,
+        sampler="dpmpp_2m_sde_gpu",
+        scheduler="normal",
+        steps=20,
+        cfg=7.0,
+        denoise=1.0,
+        seed=-1,
+        min_steps=None,
+    ):
+        self.sample_count += steps
```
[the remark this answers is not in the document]
Danamir: I see. Done !

Danamir:
I removed the use advanced option, and set the `ksampler_advanced` by default. This leaves the `sampler_steps_upscaling` only use to be in `upscale_tiled()` method. It could maybe be replaced by a simple ratio.

ai_diffusion/comfyworkflow.py
```diff
@@ -94,6 +94,45 @@ def ksampler(
...
+
+        start_at_step = round(steps*(1-denoise))
+        if min_steps and steps - start_at_step < min_steps:
+            start_at_step = math.floor(steps * 1/denoise - steps)
```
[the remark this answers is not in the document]
Danamir: I did all my tests for LCM mode, with `min_steps==steps` . This parameter should maybe be renamed to `fixed_steps`.

ai_diffusion/comfyworkflow.py
```diff
@@ -94,6 +94,45 @@ def ksampler(
...
+        positive: Output,
+        negative: Output,
+        latent_image: Output,
+        sampler="dpmpp_2m_sde_gpu",
+        scheduler="normal",
+        steps=20,
+        cfg=7.0,
+        denoise=1.0,
+        seed=-1,
+        min_steps=None,
```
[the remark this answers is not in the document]
Danamir: See the comment above, if we replace this by `fixed_steps` it makes sense to leave it as `None` to leave the computation to be done anyway. After some more testing, I did not see a case where `min_steps` would really be useful.

ai_diffusion/comfyworkflow.py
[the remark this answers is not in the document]
Danamir: `fixed_steps` could also simply be a `bool`, forcing the steps number.

Danamir:
The `start_at_step` calculation is now done in a new `_apply_strength` method to be called after `_sampler_params` in methods using the strength parameter. It alters the params dict to add `start_at_step` parameter. If the live mode is enabled, the `fixed_steps` parameter enable the correct calculation.

ai_diffusion/workflow.py
```diff
@@ -179,6 +179,20 @@ def _sampler_params(
     return params
 
 
+def _apply_strength(strength: float, params: dict, fixed_steps: bool = False) -> dict[str, Any]:
+    steps = params['steps']
+    start_at_step = round(steps*(1-strength))
```
Acly, line 184: I'd still like to see this clamped somewhere so that at least one step runs
eg. 10 steps @ 4% strength should not result in `steps==start_at_step` an no sampling happening

ai_diffusion/workflow.py
```diff
@@ -179,6 +179,20 @@ def _sampler_params(
     return params
 
 
+def _apply_strength(strength: float, params: dict, fixed_steps: bool = False) -> dict[str, Any]:
```
Acly, line 182: Hm, I'm not happy with this dict as in-out-parameter. Maybe keep the function, but return `tuple[int, int]`
```
(strength, steps, fixed_steps) -> (steps, start_at_step)
```
Then call it directly in `_sampler_params`. Maybe my suggestion with 2 functions was confusing, I didn't actually mean to separate these at the call site, just thought the step computation made a nice logical unit.

Acly:
> This leaves the `sampler_steps_upscaling` only use to be in `upscale_tiled()` method. It could maybe be replaced by a simple ratio.

I agree, let's remove it. `upscale_tiled` needs to be revisited anyway (ultimate sd upscale is a bit ugly), and it only used the upscale sample count because it was already there.

ai_diffusion/workflow.py
Danamir: I'll have pass the `strength` and `fixed_steps` parameters to `_sampler_params` if that's alright.

Danamir: Done.

ai_diffusion/workflow.py
Danamir: Done, and the calculation code is now somewhat more readable.

Danamir:
> I agree, let's remove it.

Done, `sampler_steps_upscaling` is out. I also removed it from the default styles JSON files.

ai_diffusion/style.py
```diff
@@ -210,7 +203,7 @@ def get_sampler_config(self, is_upscaling=False, is_live=False):
         if is_live:
             return SamplerConfig(self.live_sampler, self.live_sampler_steps, self.live_cfg_scale)
         if is_upscaling:
```
Acly, line 205: this branch (and the parameter) are no longer useful now

ai_diffusion/workflow.py
```diff
@@ -150,7 +150,7 @@ def __init__(self, is_active=False, strength=0.3, seed=None):
 
 
 def _sampler_params(
-    style: Style, clip_vision=False, upscale=False, live=LiveParams()
+    style: Style, clip_vision=False, upscale=False, live=LiveParams(), strength: float = None
```
Acly, line 153: * `strength: float = None` won't type-check - but it's also unclear why default is `None`, it can just be `1.0`?
* maybe rename `upscale=False` to `advanced=True` to indicate that it will create params for advanced sampler vs regular sampler. It doesn't have anything directly to do with upscale anymore.

Acly (commented):
> Done, sampler_steps_upscaling is out. I also removed it from the default styles JSON files.

Thanks, looks good! I have 2 more nitpicks (see comments) otherwise it's ready

ai_diffusion/workflow.py
Danamir: Done. You're right, its clearer this way.

ai_diffusion/style.py
Danamir: Yep, removed.

Danamir:
It should be all good now. See any more changes needed ?

Acly:
Yea it's ready and I haven't forgotten :) Just taking some time to finish UI event refactor.

Danamir:
Ok, great !

Acly:
Merged in https://github.com/Acly/krita-ai-diffusion/commit/50a2eef62a7de9751316fc92ec3513856da4ae84

I did a rebase and took the liberty to squash, it seems like GitHub doesnt allow to mark this merged though :\

PR #136, opened 2023-11-27
Added clip skip to generation and settings

Fezinaru:
I've gone through and added clip skip to both the generation and to the style settings. I went with Auto1111's style of showing the clip skip and went with a slider with range 1 to 12.  Default value is set to 1.

image


Acly:
Thank you for the PR!

Issue #138, opened 2023-11-27
ERROR:root:Failed to validate prompt for output 12:  ERROR:root:* IPAdapterApply 9:


iconnoci:
ERROR:root:Failed to validate prompt for output 12:
ERROR:root:* IPAdapterApply 9:
ERROR:root:  - Required input is missing: start_at
ERROR:root:  - Required input is missing: end_at
ERROR:root:Output will be ignored
invalid prompt: {'type': 'prompt_outputs_failed_validation', 'message': 'Prompt outputs failed validation', 'details': '', 'extra_info': {}}

![QQ截图20231127140548](https://github.com/Acly/krita-ai-diffusion/assets/22522962/88ecae6c-3a8b-4899-ba26-1b00cd46cf61)

[krita-ai-diffusion] 1.8.1
krita 5.2.1
comfy ui  SHA-1: 39e75862b248a20e8233ccee743ba5b2e977cdcf
ComfyUI_IPAdapter_plus SHA-1: a1ddc8b729d360f86cc44fc98ec0b10c94f1747b

Danamir:
It seems the latest `IPAdapterApply` node takes two new arguments in parameter. I didn't have those, so I updated my installation and two new lines appeared : 

![image](https://github.com/Acly/krita-ai-diffusion/assets/15424198/2bf87904-84bf-46dd-bb98-0875a3037811)

PR #140, opened 2023-11-27
Support new IPAdapterApply parameters

Danamir:
The `IPAdapterApply` node from https://github.com/cubiq/ComfyUI_IPAdapter_plus has two new parameters : `start_at` default `0.0` and `end_at` default `1.0` .

image

This fixes the `apply_ip_adapter` method by adding a `end_at` parameter, set to `control.end` or default `1.0` .

In the control UI we can also display the "end at" setting available to other control modes, to allow a manual setting.

image

### DIFF
--- a/ai_diffusion/client.py
+++ b/ai_diffusion/client.py
@@ -126,6 +126,7 @@ class Client:
     clip_vision_model: str
     ip_adapter_model: dict[SDVersion, str | None]
     ip_adapter_has_weight_type = False
+    ip_adapter_has_start = False
     lcm_model: dict[SDVersion, str | None]
     supported_sd_versions: list[SDVersion]
     device_info: DeviceInfo
@@ -166,6 +167,9 @@ class Client:
         }
         client.ip_adapter_has_weight_type = (
             "weight_type" in nodes["IPAdapterApply"]["input"]["required"]
+        )
+        client.ip_adapter_has_start = (
+            "start_at" in nodes["IPAdapterApply"]["input"]["required"]
         )
 
         # Retrieve upscale models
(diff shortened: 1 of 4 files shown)


Danamir:
This PR should fix #138 .

Danamir:
This was tested on my manual installation of ComfyUI . I suppose the plugin handled version of ComfyUI always get the latest version of the external nodes. But if this is not the case, this PR may have unwanted effects.

Acly:
>  I suppose the plugin handled version of ComfyUI always get the latest version of the external nodes.

It doesn't, to avoid situation where IPAdapter decides to add some new parameters and break stuff :)
The versions are fixed in `resources.py`. 
* Changing the `version` at the top of this file will trigger a server Upgrade for users with a manage installation
* Then I still have to take care of the docker image

Previously this happened with `weight_type` parameter, and I implemented it in  a way that it was compatible with both old & new version of the extension, at least for a while. But it was easy because I didn't actually use the parameter, I wouldn't do it if the code becomes more complex.

Danamir:
Ah, that's wise. 

So this fix is only needed for a custom install. I guess I'll have to check the "Connection / external server" setting to keep things relatively clean.

Acly:
> I guess I'll have to check the "Connection / external server" setting to keep things relatively clean.

I'd do it same as for `weight_type`: check the node itself for the start/end parameter when connecting. That will work independent of managed/custom, so also for users with custom install but with an old version. (At least for a "migration" period, code can be simplified after a while)

I'm in favour of fixing it that way for now, and do an upgrade of the managed install a bit later, perhaps collecting some other dependency updates.

Danamir:
That should do it.

Acly:
Yup, thank you

gxground:
but when i add controlnet ,i cannot find ipadapter in the list

Issue #148, opened 2023-11-28
FEATURE REQUEST: Menu support for subfolders


Duemellon:
Can you add menu support for subfolders for the browse actions for Loras & Checkpoints, etc.?

To be more specific:
I have my Loras in subfolders, like StyleCore, Person, etc. when I go to [add] a Lora could you add the ability for it to use the subfolder structure? So when "add" is clicked it presents something like
Novelty >
Person >
Stylecore >
Quality >

so when I then go to the added expanding fly-out menu it would show:
Novelty > Action Figure
(novelty) Stickers
(novelty) X-Ray

etc.? I have a few hundreds of loras now, so navigating it through the current method is a bit laborious

Danamir:
Ohh, great idea, I have the same exact problem.

I'm doing a PR right now : 

![image](https://github.com/Acly/krita-ai-diffusion/assets/15424198/dc38cf1c-aa9c-4071-a52f-e0da6e77b541)

Acly:
Added in [1.9.0](https://github.com/Acly/krita-ai-diffusion/releases/tag/v1.9.0)

PR #150, opened 2023-11-29
Use structured menu for styles LoRA selector

Danamir:
Replacing the `QComboBox` of the styles LoRA selector by a structured `QMenu`, to allow an easier selection with a complex file structure :  

image

NB : The method `get_path_dict(paths)` to transform a list of paths into a dict was found here : https://stackoverflow.com/questions/58916584/convert-list-of-paths-to-dictionary-in-python/58917078#58917078

This solves #148

### DIFF
--- a/ai_diffusion/ui/settings.py
+++ b/ai_diffusion/ui/settings.py
@@ -1,6 +1,10 @@
 from __future__ import annotations
+
+import functools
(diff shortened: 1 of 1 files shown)


Danamir:
Yeah, easy peasy. 😬

PR #171, opened 2023-12-02
Use arrow up/arrow down key to adjust attention weight in prompt input box

### LINKED ISSUE #170: [Feature Request] Use Up/Down arrow key to increase/decrease prompt keyword weight
It is a very handy A1111 feature that is missing in this extension right now. I can take a look and try to port the feature from A1111.

huchenlei:
Closes #170.

ezgif-4-8a57da64aa

Note:
There is a behaviour difference in multi vs single line widget, as there is a bug on `setSelection` in single line, which sometimes set the selection end to end of text when correct start/end pair are passed. Currently behaviour is multi widget creates a selection, while single widget moves the cursor.


ai_diffusion/attention_edit.py
```diff
@@ -0,0 +1,148 @@
...
+    while end < len(text) and text[end] not in delimiters:
+        end += 1
+
+    return start, end
+
+
+def select_on_cursor_pos(text: str, cursor_pos: int) -> Tuple[int, int]:
+    """Return a range in the text based on the cursor_position."""
+    return (
+        select_current_parenthesis_block(text, cursor_pos, "(", ")")
```
Acly, line 54: as far as I know, only `( )` parenthesis work in comfy

ai_diffusion/styles/default.json
```diff
@@ -0,0 +1,17 @@
+{
```
Acly, line 1: accidental commit

huchenlei:
How about making the trigger ctrl+arrow keys? I think A1111 does that in a multi-line textbox.

Acly:
> How about making the trigger ctrl+arrow keys? I think A1111 does that in a multi-line textbox.

That's also good. I just checked and it's Ctrl in Comfy's UI too, so likely people are used to it.

huchenlei:
Changed to Ctrl + arrow key binding. I think the multi-line textbox now has no unexpected behaviours.

ai_diffusion/styles/default.json
huchenlei: File removed. Should we ignore the styles directory?

Acly:
👍 Looks good

PR #198, opened 2023-12-08
Parse LoRA from prompt

Danamir:
This adds the capacity to parse LoRA directly from the prompt, using the Auto1111 webui syntax : `<lora:name:strength>`.

The lora name is searched in all the available lora models in the ComfyUI installation, by case insensitive filename only. No path, no extension, to keep the prompt light.

The strength is optional, replaced by `1.0` if absent.

Once applied the whole `<lora*>` is removed from the prompt before applying the conditioning. 

For example the prompt `photography of a bird <lora:add-detail-xl:1.0>` auto loads the matching LoRA `LoRA\5. SDXL\Style\Add-detail-XL.safetensors` found on my installation : 

image

If the LoRA is not found in the available models, an error is returned :  

image

Same concept if an invalid strength is given : 

image

I find it more responsive to display the error directly in the krita interface, but it could be replaced by a log line only in `client.log`

### DIFF
--- a/ai_diffusion/workflow.py
+++ b/ai_diffusion/workflow.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 import math
-import os.path
 import re
+from pathlib import Path
 from typing import Any, List, NamedTuple, Optional
 
 from .image import Bounds, Extent, Image, Mask
@@ -209,8 +209,7 @@ def _parse_loras(client: Client, prompt: str) -> list[dict[str, str | float]]:
         lora_name = ""
 
         for client_lora in client.lora_models:
-            _, lora_filename = os.path.split(client_lora)
-            lora_filename, _ = os.path.splitext(lora_filename)
+            lora_filename = Path(client_lora).stem
             if match[0].lower() == lora_filename.lower():
                 lora_name = client_lora
 
--- a/ai_diffusion/workflow.py
+++ b/ai_diffusion/workflow.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 import math
+import os.path
+import re
 from typing import Any, List, NamedTuple, Optional
 
 from .image import Bounds, Extent, Image, Mask
@@ -9,6 +11,14 @@ from .resources import ControlMode, SDVersion
 from .settings import settings
 from .comfyworkflow import ComfyWorkflow, Output
 from .util import client_logger as log
+
+
+_pattern_lora = re.compile(r"\s*<lora:([^:<>]+)(?::([^:<>]*))?>\s*", re.IGNORECASE)
+
+
+class LoraException(Exception):
+    def __init__(self, msg):
+        super().__init__(msg)
 
 
 class ScaledExtent(NamedTuple):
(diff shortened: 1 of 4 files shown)


Danamir:
The lint check outputs : 
image

But I'm not sure what lines needs to be reformatted. I tried to apply an auto format in IntelliJ IDEA, but many lines are impacted, some not concerned by this PR. 

Any idea ?

ai_diffusion/workflow.py
```diff
@@ -1,5 +1,7 @@
 from __future__ import annotations
 import math
+import os.path
```
Acly, line 3: Prefer to use `pathlib` over `os.path`, it's already used in a lot of places

ai_diffusion/workflow.py
```diff
@@ -11,6 +13,14 @@
 from .util import client_logger as log
 
 
+_pattern_lora = re.compile(r"\s*<lora:([^:<>]+)(?::(-?[^:<>]*))?>\s*", re.IGNORECASE)
+
+
+class LoraException(Exception):
```
Acly, line 19: Just use a basic `Exception` (unless you add extra fields or use it for type filtering)

ai_diffusion/workflow.py
```diff
@@ -223,6 +267,12 @@ def load_model_with_lora(w: ComfyWorkflow, comfy: Client, style: Style, is_live=
             continue
         model, clip = w.load_lora(model, clip, lora["name"], lora["strength"], lora["strength"])
 
+    for lora in additional_loras:
```
Acly, line 270: This could be merged with the loop above, right?
```
for lora in chain(style.loras, additional_loras):
    ...
```

ai_diffusion/workflow.py
```diff
@@ -423,7 +473,9 @@ def generate(
     batch = 1 if live.is_active else batch
 
     w = ComfyWorkflow(comfy.nodes_inputs)
-    model, clip, vae = load_model_with_lora(w, comfy, style, is_live=live.is_active)
+    model, clip, vae = load_model_with_lora(
+        w, comfy, style, is_live=live.is_active, additional_loras=_parse_loras(comfy, cond.prompt)
```
Acly, line 477: Since this is repeated a lot I would prefer if the prompt was passed to `load_model_with_lora` directly, non-optional (I don't think we have a case where we don't want to support in-prompt loras?)

Then parse loras inside the function.

Acly (commented):
This looks nice to have functionality wise, I like the error handling too.

Question: since comfy uses `embedding:file` syntax for embeddings, does it make sense to support `lora:file`? It's a bit leaner, but doesn't work for filenames that contain spaces I guess.

Requests:
* Since prompt weight/strength can now be changed with Ctrl+Up/Down, it would be nice to have this for LoRA. From a user perspective I would kind of expect it to work. Maybe it's enough to add [support for <>](https://github.com/Acly/krita-ai-diffusion/pull/171/commits/b4b393b51e5354faac78359622a355d1a2cf6835) back in, or the extra colon needs to be handled.
* Please write a test for `parse_lora` (you may want to take a lora list as argument rather than `Client`)

ai_diffusion/workflow.py
Danamir: The only difference was the log output, it's not that important. I'll merge the loops.

Danamir:
> Question: since comfy uses `embedding:file` syntax for embeddings, does it make sense to support `lora:file`? It's a bit leaner, but doesn't work for filenames that contain spaces I guess.

There is the problem with LoRA names allowing spaces, and the attention edit on Ctrl+up/down which relies entirely on delimiters. (cf. below).

> Since prompt weight/strength can now be changed with Ctrl+Up/Down, it would be nice to have this for LoRA. From a user perspective I would kind of expect it to work. Maybe it's enough to add [support for <>](https://github.com/Acly/krita-ai-diffusion/pull/171/commits/b4b393b51e5354faac78359622a355d1a2cf6835) back in, or the extra colon needs to be handled.

Done. I set the lower bound to -2.0 to allow negative LoRA strengths, and patched the parser to allow strength 1.0 to be displayed for `<>` delimiters.
 
> Please write a test for `parse_lora` (you may want to take a lora list as argument rather than `Client`)

Done.

Acly:
Thanks! works very smoothly now

peterpernhofer:
Sorry to say. But in Krita-Plugin version 1.14.1 it seems. that this (very very useful) feature is missing again.
Loras are not parsed from prompt. I have to add in manually in the connection dialog settings every time, I want to use a lora

Danamir:
It's still working for me : tested with a prompt "a bird on a branch <lora:add-detail-xl:1.0>" and the option to dump the workflow I checked that the LoRA is correctly affecting the output : 

image

Can you see if the dumped workflow is working (it can be found in `%APPDATA%\krita\ai_diffusion\logs`) ? Also check the logs to see if there is a name resolution error ?

Issue #117, opened 2023-11-24
[Request] More Samplers and Scheduler selection


Apacchi88:
HI, would it be possible for you to add other samplers that ComfyUI comes with, specifically Uni_pc_bh2. Also to be able to change to a different scheduler independently , like ddim_uniform or Karras

dominichui:
yes! please add most sampler on it, same civitai sdxl turbo model needs strange samplers.

PR #233, opened 2023-12-14
Add samplers: euler, euler a

sejoung:
Added sampler, Euler, Euler a

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,6 @@ ai_diffusion/styles/*
 !ai_diffusion/styles/digital-artwork*.json
 krita_ai_diffusion*.zip
 TODO.md
+.idea
+venv
+.DS_Store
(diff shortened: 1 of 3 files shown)


Acly:
Out of interest, why would you choose Euler/EulerA over existing alternatives? What is the trade-off?

sejoung:
Because we don't use the image as they are, so we think it would be better to have a simpler image. At this situation, Euler a was the best sampler which gives me the favorite result. I think it would be good to understand this part through the attached image below. And also #117 issue seems to have an additional request in the comments

original
euler
DPM++2MSDE Karras
eulera

Acly:
Thanks, I agree it looks better in this case. It's just difficult to provide guidance on which samplers to choose, and there are a lot of options. Euler is still popular it seems.

PR #239, opened 2023-12-16
make overflowing Add Lora menus scrollable

vtvrv:
The Add Lora menu with overflow the screen if there are too many items
problem

This commit makes the menu scrollable if there are too many items.
fix

### DIFF
--- a/ai_diffusion/ui/settings.py
+++ b/ai_diffusion/ui/settings.py
@@ -299,13 +299,15 @@ class CheckBoxSetting(SettingWidget):
     def value(self, v):
         self._checkbox.setChecked(v)
 
-def _menu_width(menu) -> int:
+
+def _menu_width(menu: QMenu) -> int:
     if not menu.isEmpty():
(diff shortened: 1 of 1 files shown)


ai_diffusion/ui/settings.py
```diff
@@ -299,7 +299,15 @@ def value(self):
     def value(self, v):
         self._checkbox.setChecked(v)
 
-
+class QMenu(QMenu):
```
Acly, line 302: The sub-class isn't really needed here (and I don't like that it has the same name)
Please just keep the function, name it `_menu_width` or similar and call it directly

Acly:
This looks like a good solution for now. Please fix lint issues before merge.

vtvrv:
> This looks like a good solution for now. Please fix lint issues before merge.

@Acly Thank you for your feedback. I've updated my pull request. I'm unable to correct the Lint error. Please make any corrections.

Acly:
Thanks!

PR #272, opened 2023-12-24
Add Lora menu drag & drop

vtvrv:
This commit enables drag & drop in the add lora menu. This would be a userful feature while search is not available.
drop

This error is displayed if the lora is not in the server's lora list
error

Because I haven't found a way to get the Comfy Server's Lora folder path I'm splitting the dropped file's path and checking if it matches a relative path in the loralist.


Acly:
There are more search and filter options for Lora now. I hope it's good enough!

Sorry for late response, I actually thought I commented earlier. I didn't like the drag&drop so much because it suggests that you can drag Lora from any path, but it actually only works if its in comfyui lora model folder.

Issue #376, opened 2024-02-02
Use Case: Prompt & Parameter Exploration


daniel-richter:
I want to share some insights for a special use case: Prompt and parameter explorations. That simply means a pure txt2image workflow with 100% variation strength without any inpainting etc. to explore how different prompts and parameters affect the outcome. 

_Krita AI Diffusion_ is fantastic for trying things out on the one hand and seeing exactly what a workflow looks like on the other. That "seeing what a workflow looks like" part is especially great because it is quite easy to build custom tools on top of that. I used `mitmproxy` to see the requests to ComfyUI (it's also great that this plugin comes with a ready-to-use Docker image and a corresponding runpod template) and used them as a template in a small tool specialized for prompt exploration. It's just a small image gallery like TypeScript/Node.js application that displays generation metadata and an extended prompt field.

<img width="746" alt="Screenshot 2024-02-02 at 14 34 29" src="https://github.com/Acly/krita-ai-diffusion/assets/12428377/68d15d04-553d-415c-81c7-96be683cd7d2">

There are some things that (currently or formerly) make it difficult in Krita (resp. easier in the separate tool) to perform that specific use case:

**Pattern Expansion**
ComfyUI supports [wildcards/dynamic prompts](https://github.com/comfyanonymous/ComfyUI?tab=readme-ov-file#notes) with syntax `{wild|card|test}`, but I don't want to have the different possibilities randomly replaced but want to have a separate prompt for each value. You could queue three different prompts for each replacements, but the strength of pattern expansion is revealed when you use multiple patterns. For example `A {blue|red} {car|truck} in the {city|country}.` produces eight different prompts. 
Pattern expansion is also very useful if you try out LoRAs and want to see the effects of different weights - e.g. `<lora:myawesomelora:{0|0.2|0.4|0.6|0.8|1}>` in combination with a fixed seed.

**Seeds**
Speaking of seeds - as a fixed (or random) seed is an integral part of prompt & parameter exploration, the seed value in Krita is quite hidden in a dropdown. I caught myself playing around in Krita with a fixed seed by mistake.

**Promt/Parameter Reuse**
[...]

PR #380, opened 2024-02-04
Add queue_front parameter to enqueue method

daniel-richter:
This PR adds a selection to the dropdown to choose where to enqueue the prompt:
- at the end of the queue (default)
- at the beginning of the queue

ComfyUI provides the parameter `front:true` for `/prompt` to realize that.

image

rel #376


daniel-richter:
Maybe Ctrl+Click would be useful as an _additional_ way to trigger queue-front.
I don't see a benefit using radio buttons instead of a combo box. Maybe use save one click, but the combobox is already a working solution.

PR #497, opened 2024-03-14
Add all comfy ui samplers and schedulers

Drakosha405:
Add all comfy ui samplers and schedulers

sequence like in comfy ui but maybe we need favorites positions on top? 

Now we can easy change samplers and schedulers when we need something special for models (Like for new Lightning models). 

PS its my 1st try to add something to  addons


Drakosha405:
have same errors with styles... so need more times

Acly:
Hi, before you invest more, let's discuss direction.

Technically we could just pull all sampler/scheduler names from Comfy server and forward them. No need to list them manually. But I did not do this intentionally:
* Many samplers are outdated, or experiments, they are not useful
* Many sampler/scheduler combinations don't make sense and don't work
* Sampler choice also has impact on Lora and node structure in case of LCM

Having lots of technical options without explanation, many of which don't work together, is bad UI design and user experience. Not everybody has a working knowledge of SD samplers and schedulers.

To me, first priority is being able to easily choose between simple options. I'm open for more low level customization, but it should not replace the basic UI I think. If I want to switch to LCM or Lightning, it should be 1 click. There is no good reason for it to be more difficult.

I'd also like to understand your motivation, why do you think existing options are not enough? What is missing?

Drakosha405:
well, using the example of new versions of Lightning models. RealVisXL V4.0 recommended DPM++ SDE Karras / DPM++ SDE,
DreamShaper XL work only with DPM++ SDE Karras(recommended by author), Juggernaut XL - DPM++ SDE
but in addon we have "Lightning": "euler","sgm_uniform" yes  we have  "DPM++ SDE Karras" but in my test DPM++ SDE normal add more details to image than DPM++ SDE Karras.
so if we can get options to change scheduler people can get better results for them
PS yes all of them maybe too much. And I think your work now its not only for fast painting(open and work) community is asking for more and more features and more settings and the samplers sheet that is now present may interfere with trying new models.
It is very difficult to maintain a balance between simple and multifunctional
and thanks you for work.

Test
Test1
euler

Acly:
Yea "Lightning" sampler was for the original Lightning model & Lora. And DPM++ SDE for all the Turbo/Lightning checkpoint merges. 

Some idea for UI:
```
Sampler: [  DPM++ 2M  ]    [✓] Karras
```
with curated options for sampler:
Euler, Euler Ancestral, DPM++ SDE, DPM++ 2M, DPM++ 2M SDE, UniPC BH2, LCM +LoRA, Lightning +LoRA

It's essentially existing options, but allows to toggle Karras independently. The Karras checkbox can be disabled when a sampler is selected that requires ddim_uniform/sgm_uniform to indicate that it's ignored.

---

Alternatively there could be a kind of preset system, if you want both best of both worlds (simple, but optional full flexibility). It's more work though.

Simple UI:
```
Sampler preset: [  Default  ]
Steps: [---[]------------] 20
```
JSON config file where presets are defined (and you can add your own):
```
{
  "Default": {
    "sampler": "dpmpp_2m",
    "scheduler": "karras",
    "steps": 20,
    "cfg": 7.0,
    "lora": []
  },
  "Fast (LCM)": {
    "sampler": "lcm",
    "scheduler": "sgm_uniform",
    "steps": 6,
    "cfg": 1.8,
    "lora": ["lcm"],
    "discrete_sampling": "lcm"
  },
  ... etc
}
```
The config file just references ComyUI sampler/scheduler names directly, so you could put anything you like.

Drakosha405:
Option 2 is more time-consuming, that's for sure, but if you look into the future, it can save time if something new appears and needs to be added. it's easier to throw in another config file than to change 3+ addon files.  like the "LCM" that was added relatively recently.

but maybe someone else has an opinion on this issue.

Acly:
I merged something like the presets

image

This is the current built-in list: https://github.com/Acly/krita-ai-diffusion/blob/main/ai_diffusion/presets/samplers.json
Can be extended with custom presets in a separate file (outside the install dir so it won't be replaced when updating)

Issue #404, opened 2024-02-11
ComfyUI errors after install the core requirements since Version 1.13.1, tested also with Version 1.14.0


forest-hermit:
Logs:

```
2024-02-09 12:45:35,280 INFO Total VRAM 7940 MB, total RAM 31773 MB
2024-02-09 12:45:35,280 INFO Set vram state to: NORMAL_VRAM
2024-02-09 12:45:35,280 INFO Device: cuda:0 NVIDIA GeForce RTX 4060 Laptop GPU : cudaMallocAsync
2024-02-09 12:45:35,280 INFO VAE dtype: torch.bfloat16
2024-02-09 12:45:35,300 INFO Using pytorch cross attention
2024-02-09 12:45:36,040 INFO 
2024-02-09 12:45:36,040 ERROR Traceback (most recent call last):
  File "/home/xxxx/.var/app/org.kde.krita/data/krita/pykrita/ai_diffusion/.server/ComfyUI/main.py", line 76, in <module>
    import execution
  File "/home/xxxx/.var/app/org.kde.krita/data/krita/pykrita/ai_diffusion/.server/ComfyUI/execution.py", line 11, in <module>
    import nodes
  File "/home/xxxx/.var/app/org.kde.krita/data/krita/pykrita/ai_diffusion/.server/ComfyUI/nodes.py", line 21, in <module>
    import comfy.samplers
  File "/home/xxxx/.var/app/org.kde.krita/data/krita/pykrita/ai_diffusion/.server/ComfyUI/comfy/samplers.py", line 1, in <module>
    from .k_diffusion import sampling as k_diffusion_sampling
  File "/home/xxxx/.var/app/org.kde.krita/data/krita/pykrita/ai_diffusion/.server/ComfyUI/comfy/k_diffusion/sampling.py", line 3, in <module>
    from scipy import integrate
  File "<frozen importlib._bootstrap>", line 1075, in _handle_fromlist
  File "/home/xxxx/.var/app/org.kde.krita/data/krita/pykrita/ai_diffusion/.server/venv/lib/python3.10/site-packages/scipy/__init__.py", line 134, in __getattr__
    return _importlib.import_module(f'scipy.{name}')
  File "/usr/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
[...]

PR #522, opened 2024-03-23
Update README.md path for Flatpak version on Linux

Toxblh:
In Linux Flatpak growup, and path for pluging doesn't match with default


Toxblh:
Ok, do not merge till #404

Acly:
I removed the paths in favour of using Krita's plugin installer script. I think it's a bit easier, works in more situations, and should be fine now also for updates since user data is stored elsewhere.

PR #555, opened 2024-03-30
Add toggle for Self-Attention Guidance ComfyUI node

FeepingCreature:
Not convinced this is an unalloyed good in tests. Seems to result in sharper images when generating de-novo. Details aren't always better though. As I understand SAG, it just makes the schedule spend more time in areas that SD attends to more strongly. I think it depends on how good a grasp the model has on the situation it's generating, whether this leads to good outcomes.

Pull request is more or less copypasted from https://github.com/Acly/krita-ai-diffusion/pull/208 (thanks!)

### DIFF
--- a/ai_diffusion/api.py
+++ b/ai_diffusion/api.py
@@ -57,6 +57,7 @@ class CheckpointInput:
     loras: list[LoraInput] = field(default_factory=list)
     clip_skip: int = 0
     v_prediction_zsnr: bool = False
+    self_attention_guidance: bool = False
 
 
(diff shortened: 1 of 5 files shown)


FeepingCreature:
Does it make more sense to put the setting in the generation UI along with the seed?

Acly:
I think it fits fine, maybe not a property of the checkpoint, but a tweak/modification to it. I don't think it's something you typically toggle all the time. 

Thanks for the PR!

PR #622, opened 2024-04-16
Add Perturbed-Attention Guidance toggle. 

FeepingCreature:
See https://ku-cvlab.github.io/Perturbed-Attention-Guidance/

I worry this is following a bad precedent with the SAG PR, of which it is a copypaste. The issue is that, in testing, _once again_ it didn't really make images look unconditionally better for me, so much as _different_ - but different in a different fashion than SAG!

I mostly want to have this PR out here in case people want to play around with it.


Acly:
I tried it out some. This one is a bit more interesting. It can give some dramatic changes for the better when you don't use a prompt. Usually that's not super useful, but at lower denoise strength it can be used as a sort of "repair" to fix & clean up images without having to describe them. The question is how to integrate that in way it can be discovered rather than hidden toggle.

For regular prompted workflows it just sort of changes them a bit... more impact than SAG but yes, not always better.  Still costs 30% performance too.

FeepingCreature:
eh, closing to clean up the pr history, I'm not really using pag anymore

PR #639, opened 2024-04-21
Regional prompting / Attention couple proof of concept

Danamir:
Hello,

Using the custom node [cgem156-ComfyUI](https://github.com/laksjdjf/cgem156-ComfyUI) , I was able to have a functional regional prompting workflow for SDXL.

Here is a proof of concept in krita-ai-diffusion, using a new control layer type "Attention", and splitting the prompt in lines and parsing for `ZONE` starting lines (or `BREAK`, `PROMPT`, `ATT`, and an optional number... TBD) and new Paint layer in Krita we are able to greatly influence the rendering.

Here is a step by step : 
 - Create a new Paint layer in Krita
 - Draw/fill the region you want to control (any color, any opacity, we use the transparency mask directly)
 - Add a new "Attention" control layer in krita-ai-diffusion interface
 - Affect the new Paint layer to the control layer
 - Repeat for any number of zones
 - Alter the prompt:
   - Add as many lines as zones starting by `ZONE ` then describe the content
   - Optionally add an extra `ZONE ` line that will be applied only to the image outside the defined zones
   - The first line of the prompt is copied at the beginning of all prompts
   - The last line of the prompt is copied at the end of all prompts

An example with a single zone : 
```
photography of
ZONE a dog
ZONE a cat
two animals, a city street in the background
```
image
image

The second ZONE is automatically affected to the cat prompt. The prompts used as attention couple are : "_photography of a dog, two animals, a city street in the background_" and "_photography of a cat, two animals, a city street in the background_".

Another example:
image
image

To do : 
 - [x] Add a control layer mode "Attention"
 - [x] Get the control layer image opacity as mask
 - [x] Split main prompt by attention zones
 - [x] Apply cond and masks to `AttentionCouple`
 - [x] Use `ETN_AttentionCouple` node
 - [x] Handle render methods
     - [x] generate
     - [x] refine
     - [x] inpaint
     - [x] refine_region
 - [x] User switch for attention couple / region prompt finder
 - [x] Load LoRA only once
 - [x] Typechecks
 - [x] Lint


Danamir:
To do : 
 - Add a button to Attention control layer to create the Krita layer
   - Maybe fill the current selection with a random color ?
 - The current way of finding the mask is kinda hacky (invert, convert to image, then convert to mask)
 - Correct lint & typecheck
 - Write tests

Danamir:
If you want to check my full ComfuUI Regional Prompting that inspired this PR : https://www.reddit.com/r/StableDiffusion/comments/1c7eaza/comfyui_easy_regional_prompting_workflow_3/

Note : The AttentionCouple node use a JS trick where you have to right click then use "add input", so a dumped workflow from krita-ai-diffusion is unusable until you connect the correct mask & cond.

Danamir:
This development be of interest to : #386 #387 #567

Danamir:
An example for #387 : 

image
image
image

illmarzini:
sorry if it s too nooby, but how do i implement this? thanks in advance.

Danamir:
Improved the coherency by subtracting the lower masks from each control layer mask. This allow to better respect the order of the control layers.

This only affect images where there is a covering between two control layers. If you put a control layer below another one covering it entirely, as intended it will have an empty mask and be ignored.

It could be an option, but from my experience it works better that way, otherwise smaller zones are frequently ignored in favor of bigger covering ones.

I tried manually computing masks intersections to combine prompts, but the difference in result is marginal, and the computation by nodes would be a nightmare.

Acly:
Hi, good to see you're still around :)

I hadn't used this node before, it looks like a good trade-off between following prompts of individual regions and still achieving some full image coherence. It felt to me like the prompt space was still limited by the combination of all region prompts: when I make them too detailed, part of the prompts are dropped quickly (more so than if I was inpainting that region exclusively).

What kind of workflow do you have in mind? My main question is, can we combine this with inpaint workflows in a synergetic way rather than creating an alternative?

By dealing with areas mainly in the prompt, we can't easily assign other control layers to specific regions. It would be great to have a UI which allows that (makes most sense for IP-Adapter, ControlNet is "area-aware" by definition but it could still be useful to compose/mask it).

Some more practical questions:
* What is the plan for 2-pass generation? Currently it doesn't work, not sure if the attention couple has to be applied to the high-res pass too, but it might...
* If there's a selection (inpaint & refine region) will that filter out areas which aren't covered by the selection automatically? That could be very useful for working on documents with fine-granular regions, since there seems to be a limit on how many of them make sense to include in one generation.
* And related to the above, if there are too many regions, can they be automatically merged? (Okay maybe this is a not-so-practical question)

Danamir:
Hi there !

I admit in my own workflows this method of rendering come kinda before the inpaint stage : when I'm still exploring the prompts and seeds. It allows to do quick changes at minimal cost (redraw slightly the masks, alter some of the zones prompts, try a new seed) and the resulting image is generally quite consistent with what I want, and with a great overall coherency because of the full rendering of the image.

When I'm happy with said image is when I switch to the inpainting / region refining stage. Then I don't use the regional prompting anymore. 

As you see the use case are quite separated, in my mind at last. It is difficult to use once doing refining passes, because it adds complexity by prompting and defining masks when what you are doing is quick alterations to different zones.

If we changed the UI to have a separated prompt by layer, and automatically calculating the attention by affected repaint region, it could maybe also be useful in the refining stage ; but the complexity of code would be an order of magnitude higher, for a somewhat unclear gain compared to multiple inpaintings.

> What kind of workflow do you have in mind? My main question is, can we combine this with inpaint workflows in a synergetic way rather than creating an alternative?

Weeeeell now that you say it like that, I was kinda creating an alternative to inpainting for quick iterations. 😅

> when I make them too detailed, part of the prompts are dropped quickly (more so than if I was inpainting that region exclusively)

I did not feel this effect, but it may be subjective : by splitting the prompt in many smaller zone prompts I feel the effect of each is stronger of course. But I'm always a very cautious prompter, trying to get an effect with the minimum number of tokens. Thus I may not be the best judge of prompt complexity.

> What is the plan for 2-pass generation? Currently it doesn't work, not sure if the attention couple has to be applied to the high-res pass too, but it might

I successfully added an upscale pass to my ComfyUI worfklow : [Danamir Regional Prompting v12.json](https://drive.google.com/file/d/1bj9__rwep6nH-FC5EG6Yhl7ggX4e0LoQ/view?usp=sharing) , but I had to resize each mask individually to keep the zones attention on the correct upscaled zone. Should be the same here, scaling the masks should suffice in theory. When I tried without attention on the second pass, the results were definitely lacking details, so there is really a gain by applying it to the first and second pass.

> If there's a selection (inpaint & refine region) will that filter out areas which aren't covered by the selection automatically? That could be very useful for working on documents with fine-granular regions, since there seems to be a limit on how many of them make sense to include in one generation.

For now I was planning to leave the selection rendering alone, because of the single prompt UI. As I was saying at above, the interface and workflows would have to be quite complex to keep those same regions when inpainting. 

> And related to the above, if there are too many regions, can they be automatically merged?

I guess we could imagine a way to merge regions, but what would be the metric to decide ? I would rather have the user merge/remove some of his regions if he feels there are too many.

Acly:
> When I'm happy with said image is when I switch to the inpainting / region refining stage. Then I don't use the regional prompting anymore.

That makes sense. What I'm interested in is not so much keeping the attention coupling up also in the later stages (although it might be nice here and there), but how they require a similar setup: different prompts are attached to different regions.

The current setup for refinement is adequate if you start with one region, tweak it until it's perfect, and then move to the next. But I think that's not commonly how people work, rather you tend to have a back and forth between subjects in the image, revisiting previous prompt setups. And it's quite tedious to redo the setup every time.

Live painting has a similar issue, you end up with short prompts that relate to one part of the image. You do a very basic sketch, then move on to another part that requires a different prompt, and so on. Once you're somewhat happy with the composition you go back and add some detail to all the parts, then do another pass for simple shading hints, etc.

Once you're convinced SD has more or less understood what you're trying to do, you might want to run it through high quality generation, find a  good seed. It would be nice if those prompts you setup for all those areas were available now to use with the attention coupling!

After you have a coherent image, you might want to refine details at low strength, focusing on specific regions. Attention coupling may no longer be needed, but those prompts you set up are still useful.

Finding good compositions is usually best at not-too-high resolution, to iterate fast. But at some point you want to increase rendering resolution. The current tiled upscaling is... very rudimentary. It doesn't use a  prompt, but it's also almost impossible to put a good one, since it has to describe each tile. If you had prompts, IP-Adapter, CN, etc. attached to regions, you could automatically select the ones appropriate for a tile, and probably get much better results.

So what I'm looking for with this is mainly an infrastructure that allows us to reuse the regional prompts (and other setup) throughout the whole lifecycle. You may not always use all of those steps and features, ... but if you didn't have to waste time redoing the whole setup you might be tempted ;)

---

After this big motivational speech, some thoughts on how it might look like:
* Attach prompts (and CN, IPA) to layer groups (if using coupling: the group defines the mask region)
* Nested groups inherit the prompt (etc.) from their parent (if using coupling: used as background prompt)
* There is one "root" group
  * If you select a child group and generate, only its region (+ some context) will be generated
  * If you select the root group, the entire image is generated (using coupling for child regions)
  * If you make a manual selection, that region is generated (maybe using coupling for affected regions only)
* It's hierarchical, you can nest as much as you like! But always the area for generation is the current selection, and the target for coupling only its direct children. This gives you a merging strategy.

I believe this approach fits how layers are used for painting. The big challenge is that diffusion is inherently layer-unaware. Any new generated image by default sits globally on top of the stack and overrides everything. Layer diffusion exists, but so far I find it somewhat clunky and incomplete. It may fit in there somehow, but I don't know how yet. So my initial approach to solve that would not rely on it, but rather use each group's mask to split new results. If the groups have a transparency mask, that would be as easy as adding results to all groups.

You'd have to manually edit them if the new generated content doesn't fit the original bounds, but it would not be hard. Or in a lot of stages, probably be okay with the mask not fitting the subject, just the general area.

Now this all sounds complex, but it's opt-in, you don't have to make your document more complicated than you need to. If it's fine with two groups, use that. If it doesn't need any, everything works as before.

Well, from a user perspective anyway. Implementation is a different matter.

Danamir:
Oooh, it's an interesting way of doing it for the future certainly. Although It's an entirely different beast to tackle than my proposition !

My current quick-and-dirty implementation has the benefit of being run in a single sampling pass, hence the lightness. In my ComfyUI workflow I'm even limited to a single LoRA, because adding one by region would alter the model multiple times, and that's not how the `AttentionCouple` custom node works. Your idea would require to develop all the logic of this node, and much more, to be able to apply each CN, mask, LoRA, IPAdapter and such, to each region. I suppose it would also necessitate many sampling passes, according to what control layers and regions are active.

As much as I would love to see this implemented, I'm not sure how much I could contribute to such a development. I was trying a few hours ago to alter the control layer's generate button to change its behavior to create a new paint layer in Krita with a nice name, filling the current selection with a flat random color at 50% opacity... and failed miserably. 😬 I can't wrap my head around how the jobs are called when you don't really want a job to be launched on the server.

Acly:
I'm not suggesting a replacement for what you're doing, I think single-pass with multiple prompts is useful (even with the lora limitation). But I would like it to use a shared UI/setup with other potential methods. Not all of this has to be implemented immediately.

I will try out some ideas for UI. Either way the code you write for backend workflow generation will be very useful. Maybe don't focus on UI too much.

> I can't wrap my head around how the jobs are called when you don't really want a job to be launched on the server.

Hm, I don't think you want anything to do with jobs in that case, they're for asynchronous background tasks. What you're trying to do can probably be done directly on button press?

Danamir:
> Maybe don't focus on UI too much.

Not a problem ! I tried to get to add a text input inside each control layer and quickly left it aside. The mix of Qt and Krita package is not my cup of tea.

> Hm, I don't think you want anything to do with jobs in that case

Ok, I was wondering about this. Everything should be able to be done in Krita, but I didn't know if the Job finishing callback part was mandatory for the UI update.

BinaryQuantumSoul:
If you find a way to add LoRa, embeddings, CN and IPA into that, it would be abolutely perfect.

AvidSpline:
Have y'all seen https://www.youtube.com/watch?v=4jq6VQHyXjg ? (Latent Vision/Mateo new IPAdapter for masked attention)

Acly:
> Have y'all seen https://www.youtube.com/watch?v=4jq6VQHyXjg ? (Latent Vision/Mateo new IPAdapter nodes to simplify masked attention).

Hadn't seen it. It looks like it's not a new method though, just conditioning masks and IPA attention masks wrapped into one node for convenience.

@Danamir I guess you must have tried conditioning masks too, I was never very happy with them and the attention couple seems to work a bit better, but really haven't done systematic test. How big is the difference?
Also, did you try attention couple + (masked) IPA? It would be good to know that it doesn't somehow interfere with each other.

Danamir:
Got some random RuntimeError from time to time that could be an issue on my system. But then when it's working, it's working pretty well. I got normal IPAdapter and FaceID working. Just be sure to send the model out from `Attention Couple` and into `IPAdapter Unified Loader`, and not the other way around.

Acly:
I created a branch: https://github.com/Acly/krita-ai-diffusion/tree/regions

It has some UI (draft) for regions which are attached to Krita group layers.  Basic setup looks like this:

regions

Can be streamlined more and missing a lot of things, but it should fill out the new [Conditioning structure](https://github.com/Acly/krita-ai-diffusion/blob/regions/ai_diffusion/workflow.py#L169-L190)  with regions. Maybe you can rebase your PR on top of that and give it a try. The `Region.mask` should load directly with `load_mask`.

Danamir:
Very neat UI. I tried it and have a few notes : 

Could it be possible to have an option to add two more prompts : 
 - A "remaining mask" (or "leftover mask") prompt affecting only the region not masked.
 - Instead of the the background prompt, separated "starting" and "ending" prompts. 

From my experience those actions can be done manually but are a great enhancement in day-to-day usage. Leaving it as an option would prevent UI cluttering. Otherwise I could still use prompt keywords to detect those sub-prompts, but it kinda defeats the new feature.

Finally, my setup of Krita crashes when I create an empty "Group Layer" manually, it only works when creating a group on an existing layer.  

image

```
-------------------

Error occurred on Monday, April 29, 2024 at 01:57:13.

krita.exe caused a Stack Overflow at location 00007FF910D5B8C7 in module python310.dll.
```

Acly:
> Could it be possible to have an option to add two more prompts :

Maybe the way the "background" is implemented right now doesn't make too much sense, it works in a hierachical way, whereas layers are first and foremost a stack.

I think if you want a prompt  that covers everything that isn't covered by a group/layer on top, you can just make another group at the bottom and fill. Maybe it needs to be more streamlined.

Technically the style prompt can already be used as the before/after prompts with the scheme "before {prompt} after". The before probably often happens to be style related, but I can see why you'd want to add some shared context after each region prompt. 

> Finally, my setup of Krita crashes when I create an empty "Group Layer" manually

Right, I fixed some stuff, probably some more, syncing regions and layers is a bit more fragile than I'd like...

Danamir:
> I think if you want a prompt that covers everything that isn't covered by a group/layer on top, you can just make another group at the bottom and fill. Maybe it needs to be more streamlined.

That is what I meant by "those actions can be done manually". At use, I found that almost every time I would have to use a "background" layer, so having it directly done by the UI simply removes a redundant step. 

>  The before probably often happens to be style related

True, it is mainly a style related feature. But it can also be useful for LoRA activation keywords : those sometimes benefits to be placed at the front of the prompt. And it can also be relevant for some composition keywords that could be ignored if placed at the end of a long prompt (ie. full body, portrait, mid shot, cowboy shot...).

Then again, I could simply detect the `{prompt}` keyword and do a simple replace. It would have to be done before applying the style to keep everything intact. 😅

Acly:
I tweaked the UI and how the "root" region works a bit. The intended use for the bottom prompt is now to be appended to all regions. I think optionally supporting `{prompt}` there would be okay too.  So if you have
* region = `region prompt`
* common (root) = `common begin {prompt} common end`
* style = `style begin {prompt} style end`

the complete prompt for region would then be `style begin common begin region prompt common end style end`

Maybe also use the "common" prompt as the background for now, still undecided on a special background text field. It's a lot of text fields. It depends on how easy it is to add regions in the end, I will try out a button for that next.

Danamir:
I like how the regions are handled. We can clearly see what control layer affect which regions. Does this mean that we won't expect control layer on multiple regions ? For the region prompting part in any case it's perfect.

I still have a weird bug if I try to use directly the "Add Group Layer" where it create a group and a layer at the same time, but the group cannot be unfolded, and the layer cannot be edited. But the "Add Paint Layer" + "Group Layer" is working as intended.

I also tried a nested group, the dedicated prompt of the nested group disappears when I click on another group, but it's generating the prompt in the workflow. I don't know if this is the expected behavior.

Questions : 

 - Is there a way to get the merged / subtracted / remaining background masks directly in Krita, by using Pillow or otherwise ; or do I try to compute it in the workflow ?
 - Do we still add an "Attention" control layer for each region we want to be affected, or do we get rid of this control layer type and affect an attention couple prompt to each region by default ?
   - The first method offers more fine control on how each region are prompted, but will necessitate to merge any sub-groups masks I suppose
   - The second method seems easier to compute (one attention couple = one mask) but could potentially leave empty prompts (which could not be a problem if the common prompt is filled)
 - For the remaining mask, would you rather have to create a group layer after all the others, then manually fill entirely the paint layer, or automatically assume that the last empty group layer is de facto the background one ? 
   - The first choice is the more logical one, but it can be weird having to fill a full layer only for having part of it effective. 
   - The second choice is faster, but maybe not intuitive.
   -  In any case the remaining mask will have to be computed, either in Krita or the workflow.

I'll try the rebase and see what I can get working.

Acly:
> I still have a weird bug if I try to use directly the "Add Group Layer" where it create a group and a layer at the same time, but the group cannot be unfolded, and the layer cannot be edited.

Strange, when I use "Add Group Layer" it just adds an empty group, no paint layer.

Some answers and updates:
* Let's keep it simple: each group is a region, each region has 1 mask and should use attention couple. No special mask control layer.
* Control layers added to a region should be applied to the region area only. Control layers added to the common/root region apply to the whole image.
* If a region has no prompt or control layers I filter it out now. This allows to still use groups for other purposes.
* Changed it so only top-level groups are considered. The order is bottom to top.
  *  Will add nesting later, my idea would be that it can be used to control complexity: nested groups are considered if you inpaint the parent.
* I've implemented that for each mask, the overlapping parts of layers above it are subtracted. Not entirely sure it's correct yet.
* I also automatically add a background region if all the region masks added up don't cover the full image. But may still change this to something more explicit. A manual background region needs to be filled.

Danamir:
> Let's keep it simple: each group is a region, each region has 1 mask and should use attention couple. No special mask control layer.

Nice, this will be easier to code. 😅

I'm wondering is there is a use case where the "group layer + mask + prompt" would be useful for a control layer, but incompatible with attention couple. I suppose there could be a checkbox in the "custom" interface do disable the attention couple entirely if needed, or maybe in the menu handling the seed and such.

> Control layers added to a region should be applied to the region area only. Control layers added to the common/root region apply to the whole image.

Perfect. I did not think to try to add a control layer to the common prompt, as I though it was only a text input. 

> I've implemented that for each mask, the overlapping parts of layers above it are subtracted. Not entirely sure it's correct yet.

I'll inspect the generated workflows to see if the computed mask is ok. The attention couple node can be peculiar on which mask it accepts. There can be no gap and no overlap of total strength > 1.0. So There may be a need for a few mask manipulation in the workflow to ensure the combined masks are complete.

Danamir:
There is a nasty bug with the way `accumulated_mask` is initialized in `RegionTree.to_api()` . The way it is done actually : 

```Python
        for region in reversed(api_regions):
            if accumulated_mask is None:
                accumulated_mask = region.mask
```

creates a pointer to the mask, so if there is only one region, not fully covering the canvas, the next portion of the code : 

```Python
            if not fully_covered:
                accumulated_mask.invert()
```

actually also inverts the first region mask ! 

I tried with a `accumulated_mask = copy(region.mask)` but it creates a shallow copy, not sufficient. Finally I simply added the mask to itself, it does nothing but it really creates a new image, preventing the later pointer overload : 

```Python
            if accumulated_mask is None:
                accumulated_mask = Image.mask_add(region.mask, region.mask)
```

There is certainly a nicer way to achieve this, but it's working. 😅

Danamir:
Here, pushed a working version with the new Regions !

Still got a strange thing where I need to convert the mask to image then to mask again, otherwise ComfyUI complaints that it's in the `[<width>, <height>]` format instead of the expected `[1, <width>, <height>]` .

It's working pretty well, it would be even better if the thumbnail was showing the full image to see the region relative position to the other. Krita has the same problem in the layers display. An example, the blue region is on the left side, and the green one on the right, but it's impossible to see at a glance in the layers : 

image

Danamir:
Updated the code to handle the calls to scale_refine_and_decode in generate. It allows for hires fix to work, correctly scaling the masks and applying the attention couple a second time. 

This also fixes the previous mask load issue as the resize use an image conversion.

Sadly, the attention couple node has an issue with ratio other than 1:1 , most of the time it fails the upscale rendering if the ratio is not square.

Danamir:
Using your method for comparing masks by subtraction then using an average pixel, I added a way to order the region prompts for the two generate method not supporting attention couple : refine region, and inpaint.

This allow to use only the region prompt under the current selection, ordering the prompt by mask coverage : 

image
The drawn regions

image image

The groups and regions

image
Resulting prompt : "blue red green"

image
Resulting prompt : "green red"

image
Resulting prompt : "blue"

Danamir:
Note concerning the LoRA loading from prompt : 

 - Bug : if the LoRA is defined in the common prompt, it is loaded multiple times
 - Feature : when using the region prompt finder in refine/inpaint, only the affected regions LoRA(s) are loaded, this could be useful

Danamir:
Finally, getting only the regional prompt with the most selection overlapping (still combined with the common prompt) seems better.

When trying to select multiple regions, too many tokens are selected and increase the risk of polluting the inpaint. More so because of the selection feathering that can extend the selected region further than necessary.

Danamir:
Fixed the multiple LoRA loading. Now the extra LoRAs are checked against the ones already loaded, either in the style definition or the other region prompts.

Danamir:
> I don't really understand the LoRA duplication problem. Shouldn't the extraction happen before any prompts are merged?

Hmm, I may have made a valid correction but for the wrong bug. After some investigation, it happens only in specific cases : 
 - The LoRA is set in the style and one of the regions, resulting in a duplicate
 - Of course if you put the LoRA activation pattern in multiple regions
 - The LoRA is set in the common prompt, and the mask are not fully covering : then an additional region is created containing the common prompt, and the LoRA is duplicated on parse

I thought it came from the prompt computations, I did not understand right away that the lora pattern was stripped from the prompt. So it was the last bug that tricked me. The correct solution would have been to simply strip the LoRA pattern when creating the additional region.

Still, the current solution prevents all sources of potential duplicates, so it's not all wrong. 😅

Danamir:
> Can we just throw attention couple on the inpaint & refine-region too? Same behavior as generate. The input to those workflows would only contain filtered regions with "sufficient" overlap.

It should be possible now that the regions contain only the portion of the mask applicable to the selection. When I started coding the prompt finder I expected to have to do more calculations, but was pleasantly surprised to find the masks already cut to size.

The main problem come from a limitation of the attention couple node, most of the resolutions are rejected if those are not multiple of 32, and even then, many ratio are not supported. I don't know if the bug is easy to resolve on their side, or if this is inherent to the attention coupling method. 

Maybe by padding the selection we could get to the correct size/ratio, but I've not found an ideal solution yet. For example, in my ComfyUI workflow I had to do the hires upscale to a multiple of 32 on both width and height, even if having to alter slightly the ration. This is far from perfect. The bug is also affecting the second pass in Krita, most of the time if the ratio is not square it will crash.

I'll try to replace the prompt finding mechanic by the attention couple one for refine region. I'll leave the Inpaint alone for now, because of the resize.

Danamir:
Yeah, sadly I got the exact same bug when doing a region refine with attention couple. Either on a small region necessitating an upscale, or on a bigger region without needing one :  

With upscale to 1024 :  
image

Without upscale : 
image

Other than this bug, it's working as intended. If I force the selection to be an accepted resolution like 1152x896 and set the grow/feather/padding to 0% the attention couple is affecting only the part of the image it should : 

image

And the masks are correctly cut to size (it was a 50% left/right mask originally) : 

image

Danamir:
I altered the code of `apply_attention` to return a modified `extent` object if necessary, containing dimensions multiple of 64 (closest dimension above or under). It also returns a boolean telling if the attention couple was affected, it could be useful to branch some code later.

Now attention couple is used directly in all modes : `generate`, `refine`, `refine_region`, and `inpaint`. For the later I had to skip the attention couple for the very last upscale after crop, but it's only at 40% strength and used only if an upscale is necessary so it should not be detrimental.

The difference of ratio because of the multiple of 64 goes unnoticed in the majority of the cases, with maybe an exception for a very narrow or tall selection, and mostly affects SD 1.5 because of the smaller original size.

The modified code checks the regions existence before altering the extent, so it should not affect the general behavior of the plugin once the regions are automatically excluded if not applicable for a selection, pending your update. In the meantime, it is applied as long as there is an existing region.

Danamir:
I'm wondering if I should hijack the "Focus" checkbox to switch from attention coupling to a single region prompt finder when refining region & inpainting. 🤔 This could be useful to prevent concept bleeding that can occurs with another close region.

Danamir:
Done.

I think this fits nicely under the same concept. But we could still split the setting into two checkboxes if needed, for example "Focus" and "Single region".

Danamir:
Fixed the hires pass that not being updated as it should, and modified the inpaint code to use attention couple on the hires pass too.

Fixed the wrong image type bug, it was an old code portion trying to subtract mask in the region prompt finder. There is still a minor inconvenience where it seems the region mask is the original mask, and not the subtracted one. So the prompt finder may chose the wrong prompt if regions masks are overlapping. To be tested.

Acly:
The focus mode is intended to restrict the prompt to the selected region, in scenarios where it would otherwise apply to a much bigger context. It basically did regional prompting with 1 region and background. 

I was just going to rewire it so it uses the new attention-couple-region mechanism to auto-create a region with the selection as mask. If there are already regions defined, I would consider the prompt "focused", and do nothing.

Filtering out gracing regions with small overlap is a good idea anyway, and if 1 region covers like 90% of the selection just use that without attention couple.

---

Regarding multiple of 32 (or 64 now?), there is already a requirement for multiple-of-8 anyway (at VAE encode, because latent is factor 8 smaller resolution). It should be possible to adapt the existing places where this is ensured and use a different value. Still a bit annoying since it will trigger resizes in more cases now. Might take a look at the node at some point to see if it can't use an internal padding or something.

Acly:
I updated the region filtering. It might still be buggy and I left some prints in there. Also the numbers probably need tweaking. It now does the following, depending on the generation bounds (full image or selection):
* Remove empty regions
* Remove regions without prompt/control
* Do a quick bounding-box check and remove regions with <10% overlap
* If mask pixels of a region cover >90% of the bounds, use that as exclusive prompt (0 regions, no attention couple)
* If mask pixels of a region cover <10% of the bounds, remove it
* If no region has >10% coverage, fall back to regular generate (0 regions, no attention couple)
* If selected regions don't have 100% coverage, create a background region with the remainder

Also added a button to quickly set up initial groups.

Apply result now sorts the result image into the affected groups and creates transparency masks if needed. Let's see how useful it is. You can do things like easily mix multiple results by toggling them per region, or adjusting the region mask to a result you like before refines. Inpaint results that only affect one region are added to that region only keeping things organized.

I will try to add more ideas/experiments before taking closer look at corner cases and issues, to get a complete picture of the whole thing before dealing with the details and messy code.

Danamir:
I merged the latest `regions` branch into the PR. It seems to work with almost no modifications. Two notes : 

 - The auto background region is created with the "background" prompt. I guess it should be empty, as this token can really alter the generation.
 - When doing a refine with a small enough selection that only one region would be used, and said selection is on the background, it uses the first region instead of refining the background.

Danamir:
Also, the new way of computing the regions seems to mess with the `{prompt}` token parsing in the common prompt.

Acly:
> A simple positive=parent.prompt is sufficient at this point.

Hm are you sure? The initial result state is returned in cases where there are no regions at all, in that case the `{prompt}` should already be "merged away" as it won't happen later. Or so I thought, not sure.

Issue might be that `result.prompt` is reused later, where perhaps `parent.prompt` should be used.

Anyway, will definitely revisit that code. One thing I still want to try out before doing some clean up, is a more explicit way of assigning regions to layers. Maybe allow layers other than groups to be regions too, with destructive updates. But it shouldn't require any extra steps.

Having multiple layers linked to the same region prompt could also be useful. Like I tried to draw/prompt something, then I want it to appear more than once in the image, so intuitively I copy/paste the layer. Would be cool if the copies are their own region following the same prompt. Will be hard to detect automatically though.

Danamir:
> in that case the {prompt} should already be "merged away" as it won't happen later.

The `merge_prompt` is used later in `encode_attention_text_prompt` in my code, and expects the `{prompt}` to be available. I use it with `region.positive` and `cond.positive` variables.

Otherwise in `RegionTree.to_api`, `merge_prompt` is used once in the `coverage > 0.9` case, and once when initializing the positive prompt. But using it with `""` as first parameter is a sure way to replace `{prompt}` by nothing. 😅

I also have an unexpected behavior where transparency masks are added to the regions, and strange things then occurs to the rendering. I had some cases where the paint layer color was used as basis for the attention couple when refining the entire image.

Acly:
> But using it with "" as first parameter is a sure way to replace {prompt} by nothing.

That was the point. For cases where no regions are used the placeholder makes no sense.

Regarding transparency masks. You can indicate your region directly with a transparency mask. If you use a paint layer, the transparency mask is created automatically on apply (it is required for correct blending). You have to edit the mask to extend or shrink the region.

Transparency masks are also the only way to make the region bigger after generating. If the alpha was applied directly to the generated result you wouldn't be able to "unerase" it. Transparency mask allows that.

Danamir:
> That was the point. For cases where no regions are used the placeholder makes no sense.

I see your point, it is correct if we remove the prompt merging from `workflow.apply_attention` to do it entirely in the `RegionTree.to_api` method.

I simplified the apply attention method by removing prompts computations and unnecessary negative prompt clip encoding. Then I altered the to api method to merge the region positive prompts with the root positive. This should be more robust.

Danamir:
I had a case where my background region was < 10% and being ignored, even when it had enough of an impact when used previously. I would propose the region cutoff to be set at 5% coverage.

Danamir:
The Conditioning input from apply_attention do not need to be in/out anymore. This simplifies the code a little bit.

Acly:
So I was a bit distracted, but finally finished the changes I wanted to try. Just tested it briefly with this PR merged in. Theres some small conflicts due to the changes in to_api, I mostly applied your changes cirectly in the regions branch now.

Regions are now independent from group layers:
* Creating  a group doesn't automatically create a region
* Removing a group doesn't remove a linked region
* A region is created with the "Add region" button
* A region can be linked to / unlinked from (one or more) group layers
* A region has to be removed explicitly

To keep region setup streamlined, the "Add region" has some smart behavior:
* By default, it creates a region, and links it to the active layer
* If the active layer is already linked, it creates a new group + paint layer and links the new region to that

This essentially allows to still do: Add region, Paint mask, Enter prompt, Add region, Paint mask, Enter prompt, etc. without any superflous clicks. But it doesn't hijack groups meant for other purposes, and allows a lot more flexibility.

The second big change is the addition of "destructive apply". This is a simplified workflow meant for live mode:
* Regions can also be linked to paint layers directly (no group)
* Applying a result in Live _replaces_ the content of the active layer
* This allows to maintain a simple 1:1 relationship between layers and groups and iterate quickly
* You can take your simple region setup to Generate mode and use it with Attention couple. On apply, all simple region layers will be _upgraded_ to groups (non-destructive apply, new results are inserted as layers)
* Live mode will use the non-destructive apply if a region is linked to a group, so you can go back and forth

There's a bunch of corner cases to iron out, the implementation certainly hasn't gotten simpler. I think usage is more predictable though, because the magic is restricted to the explicit actions ("add region" and "apply"), and outside of that everything is very explicit.

Danamir:
Great stuff ! 
I did a rough merge, and it seems to be working pretty well as is.

I'll do some more testing next week.

Acly:
I did a lot of clean up on the regions branch. 

Also I created a branch for comfyui-tooling-nodes: https://github.com/Acly/comfyui-tooling-nodes/tree/preview
It includes a (more or less) copy the Attention Couple node. I made some changes:
* Input regions are appended to a list before passed to the node. That way exported prompt workflow.json have all the required connections.
* Input masks are expected to have image resolution divided by 8. I'm considering changing the entire region mask pipeline to run at 8x lower resolution, it will speed things up. The full res is never used anyway (also in original attention couple).
* Fixed the multiple-of-64 issue. Should work now with any multiple-of-8 image resolution.

Simple example workflow: [attention_couple_simple_non64.json](https://github.com/Acly/krita-ai-diffusion/files/15483985/attention_couple_simple_non64.json)

If you have time to test & integrate that would be great. I will take a look at upscale now.

Danamir:
Nice, it was such a pain to debug when the `workflow.json` were not complete because of the previous node. 😅

I did a quick test with a workflow of mine with your implementation of Attention Couple. As long as I divide the masks size by 8 it's working perfectly. I'll see if I keep the 0.0 base mask, it seems to help in some cases to fill the possible gaps in the global mask, but it needs some more testing.

Danamir:
Question : couldn't the mask divide by 8 be done inside the Attention Couple node to simplify the workflow ? If it's a linear relation there is no need to do the computations in the workflow, and this would allow to use the mask editor in the ComfyUI image loader directly.

Acly:
> I'll see if I keep the 0.0 base mask

Note that the first element in the regions list is special. It's conditioning is ignored, instead it uses whatever conditioning is passed to KSampler. Same behavior as the base_mask input in the original node.

I don't think passing  a 0 mask makes sense unless you explicitly don't want to use the conditioning that is passed to KSampler for attention at all. If that's really what you want to do, then maybe I should change it so first element in the list uses the linked conditioning like the others. That way it would work without a 0-mask as workaround, and  be more plausible too. Just not sure if it breaks anything...

 > couldn't the mask divide by 8 be done inside the Attention Couple node to simplify the workflow ?

Yes. It's equivalent to doing ScaleBy 1/8 before passing it to the Attention Couple node. The idea is that it's better to do that as early as possible: in the context of the plugin, all the mask preparation happening in the client (subtract & fill) can be done in low res, and there is no need to send eg. full 4k masks for each region over the network to a remote Comfy, if it's all downscaled in the end anyway. It's 64x more pixels, that's a lot.

Also I figured that the original node by laksjdjf is better for using directly in Comfy web UI, but now that you mention it, it's probably easy to detect if the mask is full res and optionally do the downscale in the node.

Acly:
Made the changes. Text prompts are passed to regions now. Conditioning passed to KSampler can be the same as the first region (previous behavior) or empty, or can be used to pass ControlNet etc.

image

Danamir:
> I don't think passing a 0 mask makes sense unless you explicitly don't want to use the conditioning that is passed to KSampler for attention at all

That was only when testing my own workflow, where the masks are made programmatically, and sometimes there is a gap of one pixel between two masks. Having a 0.0 mask covering all the regions seems to fill the possible gaps and prevent breakage on some fringe cases. Nothing to do with the krita plugin, where all the masks are correct, I'll get rid of the 0 mask in the plugin and see if there is any problem.

> but now that you mention it, it's probably easy to detect if the mask is full res and optionally do the downscale in the node

That would be great, but once again that was me thinking about my own workflow ; as it would allow me to get rid of the multiple-of-64 limitation. 😅 

For the plugin it makes perfect sense to send an a small as possible mask directly.

> Text prompts are passed to regions now. Conditioning passed to KSampler can be the same as the first region (previous behavior) or empty, or can be used to pass ControlNet etc.

Perfect.

Danamir:
The code is merged.

I left the masks at full size for now, removed the 0 str base mask, and got rid of the unnecessary extent scale to a multiple of 64.

Acly:
Thanks! Maybe this would be a good point to clean up this PR and merge it into regions branch.

BinaryQuantumSoul:
Can't wait to try this

Danamir:
There may be a bug when removing regions < 10% coverage. I did a test with many small regions. In those case I got this error and a black image : 

```
F:\ComfyUI\custom_nodes\comfyui-tooling-nodes\nodes.py:69: RuntimeWarning: invalid value encountered in cast
  image = Image.fromarray(np.clip(array, 0, 255).astype(np.uint8))
```

The problem goes away if I put the coverage to a very low value in the code.

Here is the faulty workflow output : [workflow-error.json](https://github.com/Acly/krita-ai-diffusion/files/15505631/workflow.json)
And a preview of the masks kept : 
image

And the working one : [workflow-ok.json](https://github.com/Acly/krita-ai-diffusion/files/15505601/workflow.json)
And its masks : 
image

It seems that in the first case the leftover mask calculated subtracted a mask not used in the final workflow.

I don't know if the problem can be avoided at in `ETN_AttentionCouple` by patching it, or if the mask computation in `process_regions` needs to be absolutely correct.

Danamir:
Got it. The culprit is the `accumulated_mask` variable initialization : 

```Python
    # Remove from each region mask any overlapping areas from regions above it.
    accumulated_mask = None
    for i in range(len(result_regions) - 1, -1, -1):
        region, job_region = result_regions[i]
        mask = region.mask
        if accumulated_mask is None:
            accumulated_mask = Image.copy(region.mask)
        else:
            mask = Image.mask_subtract(mask, accumulated_mask)
(...)
```

As it can be done on a region that will be ignored by the coverage check just below.

To be correct, `accumulated_mask` needs to be initialized only after the first valid region is found : 

```Python
    # Remove from each region mask any overlapping areas from regions above it.
    accumulated_mask = None
    for i in range(len(result_regions) - 1, -1, -1):
        region, job_region = result_regions[i]
        mask = region.mask
        if accumulated_mask is not None:
            mask = Image.mask_subtract(mask, accumulated_mask)

        coverage = mask.average()
        if coverage > 0.9:
            # Single region covers (almost) entire image, don't use regional conditioning.
            print(f"Using single region {region.positive[:10]}: coverage is {coverage}")
            result.control += region.control
            return result, [job_region]
        elif coverage < 0.1:
            # Region has less than 10% coverage, remove it.
            print(f"Skipping region {region.positive[:10]}: coverage is {coverage}")
            result_regions.pop(i)
        else:
            # Initialize accumulated_mask if needed
            if accumulated_mask is None:
                accumulated_mask = Image.copy(region.mask)

            # Accumulate mask for next region, and store modified mask.
            accumulated_mask = Image.mask_add(accumulated_mask, region.mask)
            region.mask = mask
```

Danamir:
Still, for those little regions, I would rather use a 5% coverage check. Maybe if we had this as a setting value ?

Acly:
Ah nice find, it pays off to do some stress testing... I don't mind setting coverage check to 5% (but don't really want to put it in settings).

There is still some code I think we no longer need?
* `use_single_region` - there's Generate/Refine Region which mostly covers that
* `attention_mask`, `mask_composite`, `load_image_mask`... not sure which of those are still used
* `attention` control layer
* `find_region_prompts`

Danamir:
I was doing some tests to see if I could get rid of `use_single_region`. Since the actual region detection uses the mask + feather to do its coverage detection, with very close regions I have a prompt bleeding between two joined regions. The `use_single_region` was a trick to make sure that only the most prevalent region would be used.

Here is an example : 

The regions : 
image

The rendered image with an annoyed man and a smiling girl. You can already imagine the problem, because the man was rendered a little on the girl side : 
image

When refining only the man face, he should still be annoyed (this is using the `use_single_region` trick) : 
image

And by using the standard region detection, he is somewhat happy : 
image

So I really like this option, as it offers more predictability when doing refining.

Acly:
IMO in this case it makes more sense to adjust the region mask to better match the result, then the 90% rule should take care of it. Alternatively select the region you want and switch to "Refine Region" (I don't know if it considers the selection at the moment, but that could make sense).

In any case I don't think there needs to be a  bool passed so far down and branches in the workflow? If you want a single region, only a single region should be passed to the api/workflow.

Danamir:
> in this case it makes more sense to adjust the region mask to better match the result, then the 90% rule should take care of it

This would be ideal, but I don't see how this can be done when sometimes the rendering of a region bleeds in another one. I tried with 0% growth/feathering/padding and "Selection Mask" as context, and it stills uses the two region instead of only one. 

That being said, I perfectly understand if you want to get rid of this option. I can keep it in my custom fork only. 😅

Acly:
My main concern is that the functionality is in the wrong place, it should be much earlier with the rest of the region-selection/filtering process. If you don't want regions, just don't pass them to the workflow in the first place.

Also the `model.region_only` bool does very similar thing already, just need to tweak the logic in `model.generate`  a bit to also work when there's a selection.

daniel-richter:
I don't know whether it's already important or related: I wanted to try out the PR, but only have CloudGPU available. Since cgem156-ComfyUI isn't installed in the "Stable Diffusion ComfyUI for Krita", I tried to add it as custom node. That didn't work - should it be possible to add this/a custom node to the ComfyUI instance of the CloudGPU image?

Danamir:
The last commits doesn't use the cgem156 node anymore, but the one adapted by Acly. You have to manually switch the branch of `comfyui-tooling-nodes` to `preview` to get the node.

PR #682, opened 2024-05-03
Allow "Pre-fill" use at 0 strength

Danamir:
Trying to see if the "inpaint with model" method can do a better job than Krita's "Smart Patch Tool".

This code simply allows the strength to be set at 0% (instead of min. 1%). If a region is selected, the generate button now acts as a direct call to ComfyUI `InpaintWithModel` node with `MAT_Places512_G_fp16` as model.
image

The results are different than the patch tool, but not really better.

I think this PR can safely be ignored, but it was fun nonetheless. 😅 If anyone can find another use for it...

### DIFF
--- a/ai_diffusion/api.py
+++ b/ai_diffusion/api.py
@@ -102,6 +102,7 @@ class InpaintMode(Enum):
     remove_object = 4
     replace_background = 5
     custom = 6
+    pre_fill = 7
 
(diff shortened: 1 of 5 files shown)


Danamir:
PS : I think this code messed-up the custom menu display. Bah.

PR #715, opened 2024-05-16
Websocket support when server address is https

codescv:
On some cloud services the service address of the tunnel is https://{host}.  In that case we should replace "https" with "ws" to get the correct websocket address.

### DIFF
--- a/ai_diffusion/comfy_client.py
+++ b/ai_diffusion/comfy_client.py
@@ -360,7 +360,10 @@ def parse_url(url: str):
 
 
 def websocket_url(url_http: str):
-    return url_http.replace("http", "ws", 1)
+    if url_http.startswith("https"):
+        return url_http.replace("https", "ws", 1)
+    else:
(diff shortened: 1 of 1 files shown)


Acly:
The way it currently works and is intended is to
* replace `http` with `ws` (for local use)
* replace `https` with `wss` (for cloud service with TLS)

Which service are you using that requires some mix (https but unsecure websocket connection)?

codescv:
When using cloudflared to forward the port, the web socket doesn't work
unless I specify ws.

So this may actually be a bug of cloudflared?

Acly:
Not sure I'd call it a bug, I can't find a lot of information about cloudflared support for web socket, and whether it will encrypt it by default. Maybe it has to be setup explicitly?

FeepingCreature:
I have an Apache proxy that wraps LetsEncrypt https around my local ComfyUI. WSS works fine over it.

Issue #840, opened 2024-06-19
SSL Error on MacOS


DrewWalkup:
DESCRIPTION:
When attempting to use a Custom Server running ComfyUI with https using the plugin on Mac, there is an SSL certification verification fail.

INFO:
OS: Mac
Krita version: 5.2.2
Plugin version: 1.18.1 
GPU vendor: Nvidia

CLIENT LOG:
`2024-06-18 23:10:51,886 INFO Loaded 8 sampler presets from /Users/dw/Library/Application Support/krita/pykrita/ai_diffusion/presets/samplers.json
2024-06-18 23:10:51,891 INFO Extension initialized, Version: 1.18.0, Python: 3.10.7 (main, Dec  7 2023, 11:44:20) [Clang 13.0.0 (clang-1300.0.29.30)]
2024-06-18 23:10:51,891 INFO Loading settings from /Users/dw/Library/Application Support/krita/ai_diffusion/settings.json
2024-06-18 23:10:52,596 INFO Connecting to https://comfyui.com:8888
2024-06-18 23:10:52,912 ERROR Error: Could not establish websocket connection at wss://comfyui.com:8888: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:997) using Python 3.10.7
Traceback (most recent call last):
  File "/Users/dw/Library/Application Support/krita/pykrita/ai_diffusion/comfy_client.py", line 97, in connect
    async with websockets_client.connect(f"{wsurl}/ws?clientId={client._id}"):
  File "/Users/dw/Library/Application Support/krita/pykrita/ai_diffusion/websockets/src/websockets/legacy/client.py", line 637, in __aenter__
    return await self
  File "/Users/dw/Library/Application Support/krita/pykrita/ai_diffusion/websockets/src/websockets/legacy/client.py", line 655, in __await_impl_timeout__
    return await self.__await_impl__()
  File "/Users/dw/Library/Application Support/krita/pykrita/ai_diffusion/websockets/src/websockets/legacy/client.py", line 659, in __await_impl__
    _transport, _protocol = await self._create_connection()
  File "/Applications/krita.app/Contents/Frameworks/Python.framework/Versions/Current/lib/python3.10/asyncio/base_events.py", line 1089, in create_connection
    transport, protocol = await self._create_connection_transport(
[...]

PR #841, opened 2024-06-19
Fix for Mac SSL issue #840

DrewWalkup:
This PR fixes issue #840 .

### DIFF
--- a/ai_diffusion/comfy_client.py
+++ b/ai_diffusion/comfy_client.py
@@ -25,7 +25,7 @@ from . import resources, util
 
 if util.is_macos:
     try:
-        import certifi
+        import certifi # type: ignore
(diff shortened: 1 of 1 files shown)


Acly:
Thanks, a lot of people ran into this, but I wasn't sure it would work like that.

Can you put this at the beginning of the file (below the imports) and make it Mac-only? Something like this:
```
if util.is_macos:
    import certifi
    import os

    os.environ["SSL_CERT_FILE"] = certifi.where()
```

DrewWalkup:
Had that thought last night as well. Confirmed this morning that certifi is included in Krita's Mac site-packages, but not Windows.

Latest commit tested on Mac and Windows and works as expected.

PR #858, opened 2024-06-21
Fix issue 838: Make the Strength spinbox snap to full steps, and show steps.

### LINKED ISSUE #838: Mismatch between selected Strength and history Strength
- Create a new 1024x1024 image
- Drag the Strength slider to 80%
- Click "Refine"
- The history entry for the generated image says 75%.

image

This happens with other strengths too, e.g. generating with 60% results in a history entry with 62%.

FWIW, this is with an SDXL Turbo model. Using version 1.18.1 of the plugin.

FeepingCreature:
Fixes https://github.com/Acly/krita-ai-diffusion/issues/838

You can still enter percent, but scroll/inc/dec actions will snap to the next full stepsize. That is, if a style with 20 steps is selected, it will snap to 5, 10, 15... etc. Changing the style does not change the percent (but does change the suffix ofc). Only a deliberate input makes any change to the actual value.
Screenshot_20240622_005030

Significant credit to Claude 3.5 Sonnet to pointing me in the right direction. If I just remove a few files (websocket and two more), I can actually feed it the whole remaining Krita AI Diffusion source, basically fill up half its context window, and still only pay 30c per query.

### DIFF
--- a/ai_diffusion/ui/animation.py
+++ b/ai_diffusion/ui/animation.py
@@ -145,6 +145,7 @@ class AnimationWidget(QWidget):
                 self.generate_button.clicked.connect(model.animation.generate),
             ]
             self.queue_button.model = model
+            self.strength_slider.model = model
             self.update_mode()
             self.update_target_layers()
             self.preview_area.clear()
--- a/ai_diffusion/ui/widget.py
+++ b/ai_diffusion/ui/widget.py
@@ -44,7 +44,8 @@ from ..jobs import JobState, JobKind
 from ..model import Model, Workspace, SamplingQuality
 from ..text import LoraId, edit_attention, select_on_cursor_pos
 from ..util import ensure
-from .settings import SettingsDialog
+from ..workflow import apply_strength, snap_to_percent
+from .settings import SettingsDialog, settings
 from .theme import SignalBlocker
 from . import actions, theme
 
@@ -590,7 +591,65 @@ class TextPromptWidget(QFrame):
             self._single.setCursorPosition(len(self._single.text()))
 
 
+class StrengthSnapping:
+    model: Model
+
+    def __init__(self, model: Model):
+        self.model = model
+
+    def get_steps(self) -> tuple[int, int]:
+        if self.model.workspace is Workspace.generation:
+            is_live = False
+        elif self.model.workspace is Workspace.upscaling:
+            assert False
+        elif self.model.workspace is Workspace.live:
+            is_live = True
+        elif self.model.workspace is Workspace.animation:
+            is_live = self.model.animation.sampling_quality is SamplingQuality.fast
+        else:
+            assert False, "unknown type of workspace"
+        return self.model.style.get_steps(is_live=is_live)
+
+    def nearest_percent(self, value: int) -> int | None:
+        _, max_steps = self.get_steps()
+        steps, start_at_step = self.apply_strength(value)
+        return snap_to_percent(steps, start_at_step, max_steps=max_steps)
+
+    def apply_strength(self, value: int) -> tuple[int, int]:
+        min_steps, max_steps = self.get_steps()
+        strength = value / 100
+        return apply_strength(strength, steps=max_steps, min_steps=min_steps)
+
+
+# SpinBox variant that allows manually entering strength values,
+# but snaps to model_steps on step actions (scrolling, arrows, arrow keys).
+class StrengthSpinBox(QSpinBox):
+    snapping: StrengthSnapping | None
+
+    def __init__(self, parent=None):
+        super().__init__(parent)
+        self.snapping = None
+        # for manual input
+        self.setMinimum(1)
+        self.setMaximum(100)
+
+    def stepBy(self, steps):
+        value = max(self.minimum(), min(self.maximum(), self.value() + steps))
+        if self.snapping is not None:
+            # keep going until we hit a new snap point
+            current_point = self.nearest_snap_point(self.value())
+            while self.nearest_snap_point(value) == current_point and value > 1:
+                value += 1 if steps > 0 else -1
+            value = self.nearest_snap_point(value)
+        self.setValue(value)
+
+    def nearest_snap_point(self, value: int) -> int:
+        assert self.snapping
+        return self.snapping.nearest_percent(value) or (int(value / 5) * 5)
+
+
 class StrengthWidget(QWidget):
+    _model: Model | None
     value_changed = pyqtSignal(float)
 
     def __init__(self, slider_range: tuple[int, int] = (1, 100), parent=None):
@@ -599,29 +658,51 @@ class StrengthWidget(QWidget):
         self._layout.setContentsMargins(0, 0, 0, 0)
         self.setLayout(self._layout)
 
+        self._model = None
+
         self._slider = QSlider(Qt.Orientation.Horizontal, self)
         self._slider.setMinimum(slider_range[0])
         self._slider.setMaximum(slider_range[1])
         self._slider.setSingleStep(5)
-        self._slider.valueChanged.connect(self.notify_changed)
+        self._slider.valueChanged.connect(self.slider_changed)
 
-        self._input = QSpinBox(self)
-        self._input.setMinimum(1)
-        self._input.setMaximum(100)
-        self._input.setSingleStep(5)
+        self._input = StrengthSpinBox(self)
         self._input.setPrefix("Strength: ")
         self._input.setSuffix("%")
         self._input.valueChanged.connect(self.notify_changed)
 
+        settings.changed.connect(self.update_suffix)
+
         self._layout.addWidget(self._slider)
         self._layout.addWidget(self._input)
 
+    def slider_changed(self, value: int):
+        if self._input.snapping is not None:
+            value = self._input.snapping.nearest_percent(value) or value
+        self.notify_changed(value)
+
     def notify_changed(self, value: int):
+        self.update_suffix()
         if self._slider.value() != value:
             self._slider.setValue(value)
         if self._input.value() != value:
             self._input.setValue(value)
         self.value_changed.emit(self.value)
+
+    @property
+    def model(self):
+        return self._model
+
+    @model.setter
+    def model(self, model: Model):
+        if self._model:
+            self._model.style_changed.disconnect(self.update_suffix)
+            self._model.animation.sampling_quality_changed.disconnect(self.update_suffix)
+        self._model = model
+        self._model.style_changed.connect(self.update_suffix)
+        self._model.animation.sampling_quality_changed.connect(self.update_suffix)
+        self._input.snapping = StrengthSnapping(self._model)
+        self.update_suffix()
 
     @property
     def value(self):
@@ -631,8 +712,17 @@ class StrengthWidget(QWidget):
     def value(self, value: float):
         if value == self.value:
             return
-        self._slider.setValue(int(value * 100))
-        self._input.setValue(int(value * 100))
+        self._slider.setValue(round(value * 100))
+        self._input.setValue(round(value * 100))
+        self.update_suffix()
+
+    def update_suffix(self):
+        if not self._input.snapping or not settings.show_steps:
+            self._input.setSuffix("%")
+            return
+
+        steps, start_at_step = self._input.snapping.apply_strength(self._input.value())
+        self._input.setSuffix(f"% ({steps - start_at_step}/{steps})")
 
 
 class WorkspaceSelectWidget(QToolButton):
--- a/ai_diffusion/settings.py
+++ b/ai_diffusion/settings.py
@@ -151,6 +151,11 @@ class Settings(QObject):
         "Auto Preview", True, "Automatically preview the first generated result on the canvas"
     )
 
+    show_steps: bool
+    _show_steps = Setting(
+        "Show Steps", False, "Display the number of steps to be evaluated in the weights box."
+    )
+
     apply_behavior: ApplyBehavior
     _apply_behavior = Setting(
         "Apply Behavior",
--- a/ai_diffusion/style.py
+++ b/ai_diffusion/style.py
@@ -183,6 +183,14 @@ class Style:
         )
         return result
 
+    def get_steps(self, is_live: bool) -> tuple[int, int]:
+        sampler_name = self.live_sampler if is_live else self.sampler
+        preset = SamplerPresets.instance()[sampler_name]
+        max_steps = self.live_sampler_steps if is_live else self.sampler_steps
+        max_steps = max_steps or preset.steps
+        min_steps = min(preset.minimum_steps, max_steps)
+        return min_steps, max_steps
+
 
 def _map_sampler_preset(filepath: str | Path, name: str, steps: int, cfg: float):
     sampler_preset = SamplerPresets.instance().add_missing(name, steps, cfg)
--- a/ai_diffusion/ui/generation.py
+++ b/ai_diffusion/ui/generation.py
@@ -586,6 +586,7 @@ class GenerationWidget(QWidget):
             self.custom_inpaint.model = model
             self.generate_button.model = model
             self.queue_button.model = model
+            self.strength_slider.model = model
             self.history.model_ = model
             self.update_generate_button()
 
--- a/ai_diffusion/ui/live.py
+++ b/ai_diffusion/ui/live.py
@@ -198,6 +198,7 @@ class LiveWidget(QWidget):
             self.apply_layer_button.setEnabled(model.live.has_result)
             self.prompt_widget.region = model.regions
             self.region_widget.root = model.regions
+            self.strength_slider.model = model
             self.update_region()
             self.update_is_active()
             self.update_is_recording()
--- a/ai_diffusion/ui/settings.py
+++ b/ai_diffusion/ui/settings.py
@@ -387,6 +387,7 @@ class InterfaceSettings(SettingsTab):
             "show_negative_prompt", SwitchSetting(S._show_negative_prompt, ("Show", "Hide"), self)
         )
         self.add("auto_preview", SwitchSetting(S._auto_preview, parent=self))
+        self.add("show_steps", SwitchSetting(S._show_steps, parent=self))
         self.add("new_seed_after_apply", SwitchSetting(S._new_seed_after_apply, parent=self))
         self.add("debug_dump_workflow", SwitchSetting(S._debug_dump_workflow, parent=self))
 
--- a/ai_diffusion/workflow.py
+++ b/ai_diffusion/workflow.py
@@ -35,21 +35,20 @@ def generate_seed():
 def _sampling_from_style(style: Style, strength: float, is_live: bool):
     sampler_name = style.live_sampler if is_live else style.sampler
     cfg = style.live_cfg_scale if is_live else style.cfg_scale
-    total_steps = style.live_sampler_steps if is_live else style.sampler_steps
+    min_steps, max_steps = style.get_steps(is_live=is_live)
     preset = SamplerPresets.instance()[sampler_name]
     result = SamplingInput(
         sampler=preset.sampler,
         scheduler=preset.scheduler,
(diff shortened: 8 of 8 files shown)


FeepingCreature:
Hm, how do I ping an issue? Does it have to be in the commit description?

Acly:
While I like the idea of snap to full steps...

... I'm afraid it gets more complicated: there is a "minimum step count" (typically 4, but custom sampler presets can overwrite it). If strength% is such that resulting step count would be below the minimum, total step count is rescaled to allow matching the strength% while executing the minimum steps.

Example: total steps = 8, min steps = 4
* 50% -> steps = 0.5 * 8 = 4
* 40% -> steps = 4, total steps = 4 / 0.4 = 10
* 36% -> steps = 4, total steps = 4 / 0.36 = 11

https://github.com/Acly/krita-ai-diffusion/blob/main/ai_diffusion/workflow.py#L52

FeepingCreature:
Oooh. Wow, yeah that is evil. So if we go below 4, the total steps invisibly scale up to keep the minimum step number at 4? Does this also apply if we've configured less steps than 4 total?

Acly:
With total steps less than minimum steps, the number of steps executed will always equal total steps.

FeepingCreature:
That seems doable. Gonna take a look in the afternoon.

arcusmaximus:
For me personally, "lying" to the user (e.g. showing their selected 60% in the history even though the plugin used 62% under the hood) would honestly be fine. Snapping wise, I can imagine users would want to keep the ability to select a nice round number (purely psychologically). And showing the step count is of course useful to explain why the rounding happens, but at the same time, you could say it clutters the UI with technical details which the user doesn't really need to know about.

FeepingCreature:
I mean even just for myself, I like my inputs to have a meaningful effect. Why offer a UI if the elements you control only have a tangential relationship to the eventual effect? This is less "I should fix this bug" and more "oh so THAT'S why sometimes weird stuff happens that makes no sense!" A UI should not gaslight its user, but that's only a side effect of the idea that the model of the UI should represent the reality of the domain that it operates in. Percentages are a lie, albeit a useful lie; steps are what's real.

arcusmaximus:
You have a point, of course, but it's also useful to keep the target audience in mind - and this plugin is clearly more focused on artists who just want to make pictures, rather than technical users who want to know how everything works.

Whereas other UIs show the step count very prominently, this one hides it away in a collapsed panel somewhere in the settings. Same thing for the inpaint resolution, or the choice of whether to use an inpaint model: other UIs make you configure those manually, but this one does it all behind the scenes without telling you anything about it.

If the overarching trend is to expose less technical details than other UIs, why expose more details here? I've used A1111, Comfy, Swarm and Invoke, and none of those told me my denoise strength is getting rounded - and it was fine, because the difference was so small that it didn't matter.

FeepingCreature:
All abstractions leak. I happen to think this one leaks quite badly, as per the original bug report. It's still very much *useful* to have percentage, of course, but as I straight up didn't *realize* this corresponded to steps (obvious in hindsight), this actually clears up a lot of things that confused me at the time: like how sometimes I could change the strength in live mode and not trigger any recomputation. (Because I didn't cross a step limit...) So I do like the simplification, but I don't think it's any *less* simplified if it actually informs me what it's doing behind the scenes, especially since the percentage input, honestly, writes checks that the backend cannot cash.

arcusmaximus:
> sometimes I could change the strength in live mode and not trigger any recomputation

So have it trigger on every strength change, easy.

>  I don't think it's any less simplified if it actually informs me what it's doing behind the scenes

Having one number is quicker to read for users than three (as above, no need to clutter the UI with information that not even Comfy shows you).

Anyway, we've both made our points, so let's have Acly decide which way to go.

Acly:
I view strength% as a way to indicate "this is how much I would like the image to change". I think it's okay for the implementation to then figure out the actual step count that best matches the intent. Knowing the actual step count and denoise% might be interesting to some, but I don't see how it changes what you might do with it.

Strength% is a high-level abstraction, it is not meant to accuractly reflect how much noise is added, there is no "lying".

I'd also like to avoid having two numbers in the UI which report the same thing. Steps are less intuitive and require additional explanation for those not familiar with diffusion process.

The original bug report is easily fixed in that spirit, by showing in the history the strength% that was used to generate the result. Showing generation parameters is what the history is meant to do, that it currently exposes an implementation detail is by accident.

>  like how sometimes I could change the strength in live mode and not trigger any recomputation

This is the point that bothers me, and why I support snapping (or at least think it's worth a try). It has confused users before, and even though I'm aware of what's happening it would be nice to have a way to go the next meaningful strength.

> So have it trigger on every strength change, easy.

If you don't cross a step threshold the image will be the same. Though if there was a way to smoothly add noise independent of the step count, that might be the best solution,

arcusmaximus:
> Though if there was a way to smoothly add noise independent of the step count, that might be the best solution,

Maybe this could be useful: https://github.com/chrisgoringe/cg-noise

FeepingCreature:
Alright, sounds sensible. Is it okay if I keep the layer display behind an advanced setting, like negative prompt?

FeepingCreature:
Alright, scrolling now works correctly below min_steps. I rewrote the code; it's a bit longer but hopefully easier to follow. It now precomputes a list of spinbox values with unique step state and just continues to the next one on scroll input. This also fixes some bugs in the computation that were previously in there; the code is now more "logical" rather than a mess of math.
I've left the labels in pending a reply regarding hidden setting or if I should take it out entirely.

Acly:
I don't mind the setting as such, but it's currently quite a lot of code and connections for what it does. It would be nice if it could compute the next (or previous) step locally from current data directly in `stepBy`. That should work for snapping. But you'd need all the events for the step display - including some that are still missing I think, like changing sampler preset or total steps in style settings. Feels like it might not be worth it.

Let's remember the goal of snapping: to provide the next value with "noticeable change". This doesn't necessarily mean the next technically smallest value. There is a reason it was fixed to 5 or 10, not 1 - going from 28% to 26% isn't really useful, even if it does produce a tiny bit different image. That's why I think it can be simplified a lot by using flat 5 if below `min_steps`, and only computing next threshold on the fly otherwise.

I do like the way it's computed now because it allows reusing the same function that will then be used for the workflow, instead of having to invert the math. You can just start at current value and increment (or decrement) in a loop until you hit a threshold though, no need to precompute the array.

FeepingCreature:
I did notice the changing settings issue. I honestly don't even know how to hook that up.

If there's a helper function that *calculates* the current `steps, max_steps`, that should work for both stepping and displaying "for free". I'll try that this afternoon.

FeepingCreature:
Done, `stepBy` now steps until it hits a change in ... steps. Also the weight display is hidden behind an interface setting, Show Steps.

ai_diffusion/ui/widget.py
```diff
@@ -590,25 +593,75 @@ def move_cursor_to_end(self):
...
+        self.min_steps = None
+        self.max_steps = None
+        # for manual input
+        self.setMinimum(1)
+        self.setMaximum(100)
+
+    def step_info(self, value: int | None = None, pretty: bool = False) -> tuple[int, int] | None:
+        if self.min_steps is None or self.max_steps is None:
+            return None
+        # See workflow.py:_apply_strength
```
Acly, line 614: Better to adapt and call the function directly. The computation might change at some point.

ai_diffusion/ui/widget.py
```diff
@@ -590,25 +593,75 @@ def move_cursor_to_end(self):
...
+        strength = (value or self.value()) / 100
+        steps = self.max_steps * strength
+        # round is more accurate, but int ensures that round steps end up on round values
```
Acly, line 617: I don't really understand this comment. Why is this needed? Shouldn't it always follow workflow computation (round)?

ai_diffusion/ui/widget.py
```diff
@@ -633,6 +699,39 @@ def value(self, value: float):
...
+        return self._is_live
+
+    @is_live.setter
+    def is_live(self, is_live: bool):
+        self._is_live = is_live
+        self.update_spinbox()
+
+    def update_spinbox(self):
+        assert self.model
+        # copied from _sampling_from_style
```
Acly, line 715: I think it's worth adding a function here that can be used in both places. Maybe `style.get_steps(is_live)` which returns `(min_steps, total_steps)`?

ai_diffusion/ui/widget.py
```diff
@@ -590,25 +593,75 @@ def move_cursor_to_end(self):
...
+        self, is_live: bool | None = None, slider_range: tuple[int, int] = (1, 100), parent=None
+    ):
         super().__init__(parent)
         self._layout = QHBoxLayout()
         self._layout.setContentsMargins(0, 0, 0, 0)
         self.setLayout(self._layout)
 
+        self._is_live = is_live
+        self._model = None
+
```
Acly, line 657: You can also connect `settings.changed` event here if you want to have immediate update when switching step display.

Example handler: https://github.com/Acly/krita-ai-diffusion/blob/main/ai_diffusion/ui/region.py#L321

Acly (commented):
I just noticed this is only for spin box and I always use the slider :)

Can snap be done for slider too? Maybe by applying the logic in `valueChanged` event, I think it might just work.

ai_diffusion/ui/widget.py
FeepingCreature: The problem is: say you have ten steps. Thanks to round(), you map 45-55 to 10/20. But snapping to "45% (10/20)" would be very strange. And finding the midpoint manually would make the code awkward. So `pretty=True` changes the interval to *start* at the natural midpoint.

FeepingCreature:
> I just noticed this is only for spin box and I always use the slider :)
> 
> Can snap be done for slider too? Maybe by applying the logic in `valueChanged` event, I think it might just work.

Absolutely, but I didn't want to overload the PR. If it's fine, I'll have a look later.

ai_diffusion/ui/animation.py
```diff
@@ -139,14 +139,18 @@ def model(self, model: Model):
                 model.error_changed.connect(self.error_text.setText),
                 model.has_error_changed.connect(self.error_text.setVisible),
                 model.layers.changed.connect(self.update_target_layers),
+                model.style_changed.connect(self.strength_slider.update_suffix),
```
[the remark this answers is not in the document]
FeepingCreature: oops, removed

ai_diffusion/ui/widget.py
FeepingCreature: oh hey nice. done

ai_diffusion/ui/widget.py
FeepingCreature: done `style.get_steps`

ai_diffusion/ui/widget.py
FeepingCreature: done

ai_diffusion/ui/animation.py
```diff
@@ -139,14 +139,18 @@ def model(self, model: Model):
...
                 self.add_control_button.clicked.connect(model.regions.add_control),
                 self.prompt_textbox.activated.connect(model.animation.generate),
                 self.negative_textbox.activated.connect(model.animation.generate),
                 self.generate_button.clicked.connect(model.animation.generate),
+                self.style_select.quality_changed.connect(self.update_strength_slider_live_mode),
```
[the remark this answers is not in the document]
Acly: Yea you still have to connect/check `model.animation.sampling_quality_changed` if `workspace is Workspace.animation` - but you already pass in the model, so you can handle it inside the strength widget. It's not a big difference, just makes it more self-contained.

ai_diffusion/ui/widget.py
Acly: Okay I see what you mean. It's still super confusing if you just look at the `apply_strength` code, it says `round_up` but actually does a `ceil`. And it technically gives wrong results if you don't use it in this exact searching-the-next-threshold iterative way.

Why not loop until you find the next threshold using the original function with round. Say at 45 steps goes from 9 to 10. Then return `steps / max_steps`, which is 50.

FeepingCreature:
Aigh, slider should snap now.

FeepingCreature:
I've hit an issue. Sometimes, I can get a signal loop because the value change propagates to the widgets in every workspace, and depending on liveness they can have different snap settings, which can cycle.

Hm. I guess this is a forcing case, actually, for just doing what you said and just having the *model*, not the `is_live` flag, determine the `is_live` state in every widget. Ie. the Generation widget is also treated as `is_live` if the Animation widget is active and in fast mode.

Acly:
I think a safer way to break the signal loop is to do it explicitly somewhere. Instead of connecting `model.strength_changed` to `slider.setValue` directly, use a function, and avoid snapping model updates with `SignalBlocker`.

FeepingCreature:
Yeah but I'm actually just confused again, like I'm not understanding what's happening. Looking at the code more, I don't actually think it's *supposed* to propagate value changes between the widgets? I have no idea where the second widget object that's seeing value changes there is coming from.

Acly:
You have multiple widgets linked to the same model with bi-directional links. If the widget changes so does the model, if the model changes so do all the widgets. _Usually_ this is broken by checking if the value _actually_ changed, and not emitting a signal if it didn't. If you have inconsistent snapping that would cause a loop.

FeepingCreature:
OH RIGHT, STRENGTH IS IN THE MODEL. Thanks.

FeepingCreature:
I think I got it. It's even simpler now.

`StrengthSnapping` just stores the `model` and checks for bounds and is_live on the fly. That way every widget sees the same bounds always. Loop seems to be gone, no more weird behaviors I can see offhand

ai_diffusion/ui/animation.py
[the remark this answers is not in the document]
FeepingCreature: Shouuuuld now be handled?

FeepingCreature:
Okay, I've been trying to apply your review to use `steps/max_steps` to make the center search easier. 

Problem: `apply_strength` is unstable.

Example:

Say we have settings: min steps 3, steps 8

We start with 31%: a strength of 0.31.
In `apply_strength` with `round(5.52)` this comes out to start at step 6, which violates `min_steps`.

So we recalc `steps` as `3/0.31` = `9.67`, floor to 9, start at 6.

But now with 3/9, we reconstruct a strength of `0.33`, which is more than we started at and no longer even violates min_steps!

Acly:
If it's just the floor causing issues, it could be turned into a round (will more closely approximate the desired strength anyway).

FeepingCreature:
It's weirder.

I've had Claude 3.5 Sonnet whip me up a Python graph tester.

https://gist.github.com/FeepingCreature/1a1ad917a717a891ac2d280947003882

As far as I can tell from poking around, this instability happens **only** with steps 8, min_steps 3.

FeepingCreature:
https://gist.github.com/FeepingCreature/9923065a35129624b27c8b033dca0575 I asked Sonnet for a separate tool to just log all instabilities. Apparently (experimentally), the set of functions with the lowest errors is round twice in apply_strength, then round again in setValue. However, it still has configurations where it isn't stable. :/

Acly:
I don't think investing work to make it stable is a good idea either, it makes tweaking `apply_strength` too complicated.

I still think snap should just not do any snapping if steps==min_steps, I don't see any value in it

FeepingCreature:
Ah, if all the instability is at or below min_steps, that would fix it, right? Yeah then it makes sense to unlock the step size below that. And it'll let me use the simpler computation for the middle instead of stepping.

Okay, I've run it through the tester and it looks like so long as steps stay limited to 100, there are no stability errors if we ignore the region below min_steps.

ai_diffusion/ui/widget.py
FeepingCreature: Should be done now.

FeepingCreature:
ALRIGHT! I feel good about this version!

ai_diffusion/ui/widget.py
```diff
@@ -590,7 +591,66 @@ def move_cursor_to_end(self):
...
+            assert False, "unknown type of workspace"
+        return self.model.style.get_steps(is_live=is_live)
+
+    def nearest_percent(self, value: int) -> int | None:
+        min_steps, max_steps = self.get_steps()
+        strength = value / 100
+        steps, start_at_step = apply_strength(strength, steps=max_steps, min_steps=min_steps)
+        return snap_to_percent(steps, start_at_step, max_steps=max_steps)
+
+    def apply_strength(self, value: int) -> tuple[int, int]:
```
Acly, line 619: looks like `nearest_percent` should make use of this function

ai_diffusion/ui/widget.py
```diff
@@ -599,30 +659,52 @@ def __init__(self, slider_range: tuple[int, int] = (1, 100), parent=None):
...
         self.setLayout(self._layout)
 
+        self._model = None
+
         self._slider = QSlider(Qt.Orientation.Horizontal, self)
         self._slider.setMinimum(slider_range[0])
         self._slider.setMaximum(slider_range[1])
-        self._slider.setSingleStep(5)
-        self._slider.valueChanged.connect(self.notify_changed)
+        self._slider.setSingleStep(1)
```
Acly, line 667: Why this change? I like 5. It means scroll-wheeling over the slider changes in steps of 10, which feels nice. With singleStep=1 scroll-wheeling over the slider doesn't work anymore at all in a lot of situations (threshold interval > 5).

Technically this is still a problem when total steps is as low as 4, because in that case 10 won't be enough to jump to the next threshold, but it's a rare case and can be accepted...

ai_diffusion/ui/widget.py
FeepingCreature: Not totally clean cause we still need `max_steps`. But done.

ai_diffusion/ui/widget.py
FeepingCreature: I wanted to avoid overscrolling (two steps at a time) when for instance the natural step size is 3. But since I don't scroll-wheel the slider myself anyways, I guess I don't mind. Done.

FeepingCreature: Note that I still have the input box set to singleStep(1), cause there it'd really be weird to have "arrow-up" go two steps.

Acly: Yea it's not an issue for the spin box since it supports custom step function. The way snap is done for the slider unfortunately means if you use arrow keys or scroll wheel and don't reach the next threshold, it always snaps back to the previous value, and never changes at all.

Anyway let's hope for now this solution is good enough and nobody actually uses arrow keys.

FeepingCreature: In case I forgot to mention it: the reason I do it with `value` is that I tried overriding the mouse move event and that makes the UI so slow it becomes unusable. So I basically can't touch those methods with python which means I can't notice *why* the value changed. Maybe it has something like stepBy I can override?

FeepingCreature: But yeah probably nobody's using it.

Acly: Maybe it's possible by overriding [`QSlider.triggerAction`](https://doc.qt.io/qt-6/qabstractslider.html#triggerAction) but I haven't tried.

I will merge it later today if you don't plan further changes.

FeepingCreature: Cheers! No, far as I can tell it's good.

Acly: Done, already getting used to it :)
Thanks for all the work and I hope you didn't get burned too hard by all the Qt signal stuff!

FeepingCreature: It was definitely a handful, but I think it ended up well for it.

I worry a bit that the code seemed reasonable on the first try, and then there turned out to be many changes necessary where at the end it still seems reasonable, and even a bunch simpler, but also it turns out the exact way it looks is exactly how it *had* to look and no other way was going to work on pain of terrible crashes. But honestly, I'm gonna stop thinking about it now. :)

PR #881, opened 2024-06-26
Use SamplerCustomAdvanced node to allow new schedulers usage (including AlingYourSteps)

Danamir:
Here is an update replacing the old `KSamplerAdvanced` by the more recent additions to base ComfyUI : `SamplerCustomAdvanced`, `*Scheduler`, `SplitSigmas`, `RandomNoise`, `CFGGuider`, `KSamplerSelect`.

The functionalities are exactly the same, ie. : `steps`, `start_at_step`, `cfg`, `add_noise`, etc. But it also allow the use of various new scheduler nodes : `BasicScheduler`, `AlignYourStepsScheduler`, `GITSScheduler`, `PolyexponentialScheduler` which are not otherwise available in the schedulers list of the old sampler node.

Coincidentally this allows the usage of the new iPNDM V sampler with Align Your Steps scheduler, which works really well : 

```json
    "iPNDM V": {
        "sampler": "ipndm_v",
        "scheduler": "align_your_steps",
        "steps": 20,
        "cfg": 6.0
    }
```

### DIFF
--- a/ai_diffusion/comfy_workflow.py
+++ b/ai_diffusion/comfy_workflow.py
@@ -192,11 +192,15 @@ class ComfyWorkflow:
             noise=self.random_noise(seed),
             guider=self.cfg_guider(model, positive, negative, cfg),
             sampler=self.ksampler_select(sampler),
-            sigmas=self.split_sigmas(self.scheduler_sigmas(model, scheduler, steps, model_version), start_at_step)[1],
+            sigmas=self.split_sigmas(
+                self.scheduler_sigmas(model, scheduler, steps, model_version), start_at_step
+            )[1],
             latent_image=latent_image,
         )[1]
 
-    def scheduler_sigmas(self, model: Output, scheduler="normal", steps=20, model_version=SDVersion.sdxl):
+    def scheduler_sigmas(
+        self, model: Output, scheduler="normal", steps=20, model_version=SDVersion.sdxl
+    ):
         if scheduler in ("align_your_steps", "ays"):
             assert model_version in (SDVersion.sd15, SDVersion.sdxl)
 
--- a/ai_diffusion/comfy_workflow.py
+++ b/ai_diffusion/comfy_workflow.py
@@ -176,6 +176,7 @@ class ComfyWorkflow:
         positive: Output,
         negative: Output,
         latent_image: Output,
+        model_version: SDVersion,
         sampler="dpmpp_2m_sde_gpu",
         scheduler="normal",
         steps=20,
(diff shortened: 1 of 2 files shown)


Danamir:
This development works with my custom installation of ComfyUI, but I'm not sure the bundled version with krita-ai-diffusion has those, to be tested.

And sadly the latest ComfyUI version seems to have trouble loading the lightweights `workflow.json` generated by the plugin, significantly increasing the difficulty to debug.

Danamir:
Here is a discussion on iPNDM samplers : https://github.com/comfyanonymous/ComfyUI/discussions/3799

Acly:
I tried AYS when I tested all the distilled low-step Lora for last update, and thought it was kind of meh. Haven't tried out IPNDM yet though. And there is also CFG++ now...

Danamir:
Yeah, I would not have done the development for AYS only. But I'm guessing more and more of the future scheduler will be ported to ComfyUI as `*Scheduler` nodes, so I did this to mostly to future-proof the sampling method used. 

I always has been a fervent user of DPM++ SDE (not the 2M/3M variants) because I felt it had more coherency, responded well to LoRAs, and had less anatomy problems. Right now I'm testing iPNDM V with AYS/GITS, and for me it can safely replace DPM++ 2M/3M SDE, having zero color artifacts problems at low steps (~18). DPM++ SDE still has a little advantage overall, but is twice as slow. 

I did not see CFG++, is it implemented in base ComfyUI or as a custom node ?

Danamir:
There you go, CFG++ support. 

Note : I thought it was only usable with a cfg between `0.0` and `1.0`, but the ComfyUI implementation ranges from `1.0` to `2.0`, cf. : https://github.com/comfyanonymous/ComfyUI/discussions/3844 

I'm having good results with AYS and `1.6` cfg : 

```json
    "Euler CFG++ AYS": {
        "sampler": "euler_cfgpp",
        "scheduler": "align_your_steps",
        "steps": 20,
        "cfg": 1.6
    }
```

Danamir:
Damn, found some more doc and I could have used directly the `euler_pp` and `euler_ancestral_pp` in `BasicScheduler`. 😅  The code evolves so fast it is difficult to get everything right on the first try.

Tested : `euler_pp` corresponds to `SamplerEulerCFGpp` with `version=alternative`. The `version=regular` is much more contrasted. I'll leave it as it is is for now do do some more tests.

Danamir:
I don't know if I'm missing something, but `euler 6.0 cfg` gives the exact same result as `euler_pp 3.0 cfg++`. Same results at 20 steps and 50 steps. I don't see the point of CFG++ then...

At least the "regular" SamplerEulerCFGpp offers different results, but impossible to say if those are better or not.

Not impressed by CFG++ so far.

Acly:
Yup, not sure what euler_pp is about either. The regular SamplerEulerCFGpp works but I don't exactly find it easier to use. Anyway I guess we can keep it, just in case someone asks.

Danamir:
After some more testing, I keep coming back to GITS scheduler ! At 20 steps and lower (when refining), I find it almost always better than Karras, which was my go-to scheduler. Above 20 steps, AYS and GITS perform equally well, so in fine GITS is more versatile.

Either with DPM++ SDE for a boost of quality, DPM++ 2M for speed, or Euler A for sharpness... it's really good.

iPNDM V is not bad at all, good to have as a choice, but I would not select it as my first general sampler.

Acly:
AYS/GITS is definitely better for 12~18 steps range. I'll probably make it the default for the "Fast" preset. It uses UniPC which actually works okay with these, though iPNDM might be better.

PR #919, opened 2024-07-07
Let QPlainTextEdit handle Ctrl+Backspace

### LINKED ISSUE #798: Prevent Krita's layer fill on Ctrl-Backspace in text field
When you're typing a prompt (regular or region), change your mind and press ctrl-backspace to delete the last word, the word does not get deleted and Krita instead fills the current layer with the secondary color. This gets annoying after a few times - would it be possible to fix?

(The shortcut works as expected when e.g. renaming a layer, so at least clearing the layer doesn't seem to be entirely unavoidable. But who knows, maybe the rules are different for plugins.)

dyscorv:
Fixes #798.

### DIFF
--- a/ai_diffusion/ui/widget.py
+++ b/ai_diffusion/ui/widget.py
@@ -33,6 +33,7 @@ from PyQt5.QtGui import (
     QPainter,
     QIcon,
     QPaintEvent,
+    QKeySequence,
 )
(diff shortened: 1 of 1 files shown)


Acly:
Thanks! Qt has thought of everything...

PR #183, opened 2023-12-04
Add tag autocomplete based on danbooru dataset

huchenlei:
Closes #173.

WIP

PR #953, opened 2024-07-20
Implement Danbooru tag completion alongside the LORA completer.

FeepingCreature:
Following up on #183 , @huchenlei do you mind if I take this over? @Acly if this feature is still acceptable, what remains to be done? IMO, we need this:

- settings section for tag sources, probably best to limit it to a fixed set to start because it seems everyone has their own syntax
- off by default
- make loading better? it's hacky right now
- completer should show tags that *start* with the current text before tags that merely contain it
- sort by popularity
- (edit) autocomplete across spaces, splitting at comma in tag mode


Acly:
Fine with me. I don't think we need the download for a 2mb file, could just put it in `ai_diffusion/tags`.

For settings, use a single combo box with options:
* None
* List CSV files in `install_dir/tags`
* List CSV files in `user_dir/tags`

That way people can supply their own if someone really wants to, without having to add extra mechanism for that.

FeepingCreature:
Personally my instinct is not to put large files in git. ... Aaaand I just realized you already have LFS enabled, nevermind, I'll put it there.

Re `user_dir/tags`, I guess my problem would be what format to use. I guess the minimum would be just a flat list of tags, maybe even txt and not even csv, and show them in whatever order they're in in that file. Then people can presort if they want something special like drop rare tags.

Acly:
GitHub has strange pricing for LFS, so I'd rather not put it there. It's not that big and text compresses reasonably well. Could strip the frequency/synonyms if they're not needed.

Pre-sorted plain list of tags sounds good.

FeepingCreature:
- Added settings section in the Interface tab
- Tags are `plugin_dir/tags/*.txt, user_dir/tags/*.txt`
- I don't think Qt supports order StartsWithThenContains. Too complex, gave up.
- Tag autocomplete now includes spaces and splits at comma.

irldoggo:
image

Line 454 in settings_widget.py has a typo. From enabled to isEnabled.

Acly:
I assumed you'd switch between different tag files, not use multiple at the same time. If you want to merge on the fly keeping the counts makes sense. I don't have much of an opinion tbh.

FeepingCreature:
Most hentai models being trained these days merge as many datasets as they can. Turns out as with every other AI, more data = better. A consequence of that is that the clips understand tags from each dataset. That's why I have multiple.

I think I'll switch to the CSV format used by the previous PR. Purely so that, if we ever expand the feature to, for instance, search through aliases or show tag frequency counts, we don't have to make people recreate their custom tag files.

FeepingCreature:
@irldoggo are you sure? that's `@property def enabled` from line 112 same file, not the Qt property.

FeepingCreature:
Haha! I was talking to an AI (Opus 3.5 Sonnet) about implementing a more pretty autocomplete dropdown. I mentioned the second column was the type of tag, and completely unprompted, it threw this enum at me:

```
class TagType(Enum):
    GENERAL = 0
    CHARACTER = 1
    COPYRIGHT = 2
    ARTIST = 3
    META = 4
```
It's not *exactly* the same, but it seems suspiciously familiar. Now I'm wondering if it's remembering seeing the previous PR ... ;-)

FeepingCreature:
@irldoggo nevermind I see what you mean. There's a type error there somewhere... fixing.

FeepingCreature:
Changes:

- Tags are now CSV again. The e621 tag file has been remapped so that its tag types match the Danbooru format.
- The widget is a subclass of SettingWidget, hopefully (seemingly) fixing @irldoggo 's bug. Because all settings widgets are VBoxes, and I wanted the file list below it, the file list widget is now left dangling off the FileListWidget as an accessor and the calling code picks it up and injects it into the interface tab vbox layout. I'm finding this a lot simpler.
- The popup is a lot prettier:
- - Each row now looks like "<tag>    <source> <count>". 
- - The CSV information is actually used: the tag type determines the background color of each entry. The color scheme is taken from danbooru's tag colors.

FeepingCreature:
And I split all the tag code out into `ai_diffusion/ui/autocomplete.py`.

ai_diffusion/ui/autocomplete.py
```diff
@@ -0,0 +1,243 @@
...
+from PyQt5.QtCore import Qt, QStringListModel, QSize, QRect, QAbstractProxyModel
+
+from ..root import root
+from ..settings import settings
+from ..text import LoraId
+from ..util import ensure, plugin_dir, user_data_dir
+
+
+class TagType(Enum):
+    General = 0
```
Acly, line 18: For consistency: use lower case for enum tags

ai_diffusion/ui/autocomplete.py
```diff
@@ -0,0 +1,243 @@
...
+            TagType.Copyright: QColor(255, 255, 230),  # Light pink
+            TagType.Character: QColor(230, 255, 230),  # Light green
+            TagType.Meta: QColor(230, 255, 255),  # Light yellow
+        }
+        return colors.get(tag_type, QColor(255, 255, 255))  # Default to white
+
+
+# Ensure there's only one of these globally. It gets pretty big if we have one per widget.
+TAG_MODEL = TagListModel([])
+TAG_FILES = None
```
Acly, line 119: please rename to `_tag_model` and `_tag_files` (they're not constants and not public)

Acly (commented):
The tag list does look swanky, but only with light theme :)

I'm not so happy with the file list:
image

Lots of empty space and for some reason the checkbox borders aren't very visible in dark theme either.

Do tags need to be escaped?
* Replace `(` and `)` with `\(` and `\)`?
* Replace space with underscore?

The fract that space is not a separator makes it very confusing, especially when you auto-complete, then space, then try to add more tags/words/whatever weird things happen.

FeepingCreature:
I'm not sure what's going on with the file selector in dark mode. I mean, it's a system component, it's not like I themed it or anything. All I can tell you is it looked very janky when I put it inline.

- [ ] make colors reactive to dark/light mode

FeepingCreature:
I talked to Pony support and they say I should use comma-separated tags with spaces. So I guess I'll keep it like that.

ai_diffusion/ui/autocomplete.py
```diff
@@ -0,0 +1,243 @@
...
+    def _background_color(self, tag_type):
+        # danbooru tag color scheme
+        colors = {
+            TagType.General: QColor(230, 230, 255),  # Light blue
+            TagType.Artist: QColor(255, 230, 230),  # Light red
+            TagType.Copyright: QColor(255, 255, 230),  # Light pink
+            TagType.Character: QColor(230, 255, 230),  # Light green
+            TagType.Meta: QColor(230, 255, 255),  # Light yellow
+        }
+        return colors.get(tag_type, QColor(255, 255, 255))  # Default to white
```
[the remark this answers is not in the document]
FeepingCreature: tag color should be reactive to theme change now

ai_diffusion/ui/autocomplete.py
FeepingCreature: done

ai_diffusion/ui/autocomplete.py
FeepingCreature: done

Acly:
I only mentioned underscores because I remember seeing that a lot, but that was a while ago and probably for SD1.5 novelai stuff. Spaces in tags are fine otherwise. I still think the auto-complete is more broken than it needs to be when you _don't_ immediately type a comma after completion. Sometimes you might mix tags and text?

Maybe it's possible to make it stop suggesting the tag that's already been completed. Not sure if it's difficult/worth it, but maybe worth a thought.

FeepingCreature:
Something like... if there's a perfect match, close the popup? Idk. I always put commas... not sure if it's good. Might be good to get some more voices about how people use the prompt?

Acly:
Also have no idea what's up with that list box. But I think list widget is not really needed anyway.

Here is just some checkboxes:
image

Can also be put below the header, I don't actually mind that (space for more files), just the list box was kinda massive.
```python
    def __init__(self, setting: Setting, files: list[str], parent=None):
        super().__init__(setting, parent)

        self._list_items: list[tuple[str, QCheckBox]] = []
        self._list_layout = QHBoxLayout()
        self._list_widget = QWidget(self)
        self._list_widget.setLayout(self._list_layout)
        self._layout.addWidget(self._list_widget)
        self._widget = self._list_widget

        self._set_files(files)

    # def reset_files(self, files: list[str]):      UNCHANGED

    def _set_files(self, files: list[str]):
        self._files = files
        self._list_widget.children().clear()
        self._list_items.clear()
        for file_ in self._files:
            checkbox = QCheckBox(file_, self)
            checkbox.toggled.connect(self._notify_value_changed)
            self._list_layout.addWidget(checkbox)
            self._list_items.append((file_, checkbox))

    @property
    def value(self):
        return [file for file, checkbox in self._list_items if checkbox.isChecked()]

    @value.setter
    def value(self, v):
        for file, checkbox in self._list_items:
            checkbox.setChecked(file in v)
```

FeepingCreature:
Added your changes. I'm not entirely sure it's a good idea: if you add more files, that list can start looking weird. Then again, will people really add more tag files ~ever? Or more than one ~ever? -- And if that ever happens, we can still go back to a bigger list. I guess my main problem is that the toggle list doesn't really suggest that the file list can be extended, but that's such a niche usecase that I think it probably doesn't matter.

Acly:
Yea I also rather optimize for the 99%-case. If someone wants to use their own tags it's possible, and if someone comes with an extensive tag library spread over several files... it's not hard to add a scrollbar to the widget later.

I can merge later today if you don't have any other changes in mind.

FeepingCreature:
Unless I'm forgetting something, seems good to me.

I noticed some weirdness with tag completions maybe starting at an off-by-one wrong offset sometimes, but I can fix that in a follow-up if I actually figure out what causes it.

Acly:
Did you add escaping for brackets? We talked about space/underscore, but the `()` defintely need to be escaped since they are interpreted by Comfy as weight adjustment.

FeepingCreature:
Ah! No, that's not in yet. Hold the presses... I'll have to look into how comfy does it. The brackets in danbooru tags always start in the middle of a tag, theoretically comfy should know that can't be a weight marker. I'll have a look at its parsing logic after work.

Acly:
I think you just write `\(` and `\)`, Comfy will pass it through as `(` and `)`.

FeepingCreature:
That's my default plan yeah, I just want a second look. I'm off work in a two hours, let me just take a look then.

Acly:
sure, there's no rush ^^

FeepingCreature:
Ugh. Update: Okay, first of all you're right that it parses a `()` tree anywhere it appears in the prompt. Second, yeah it does escaping... by replacing `\(` with `\0\1` before parsing.

Maybe a PR for Comfy with a proper parser is in order.

FeepingCreature:
Forgot to say: Parentheses are escaped now. It's a bit iffy because the autocompletion search doesn't undo the escaping at the moment, so once you accept the completion the search doesn't work anymore. I'll fix that after work again.

Acly:
Sounds good. Another small request: could you please escape the escaping `\`, ie. `"\\("` and `"\\)"`

FeepingCreature:
... I don't understand. Why do we need that?

FeepingCreature:
- Escaped backslashes in Python strings
- Unescape backslashes when searching the completion
- Fixed the inserted completion being at a wrong offset when parentheses were in play

Acly:
Merged, but without e621 for now. Sorry for the ninja-edit, I realized after the merge that people will generally not know what e621 stands for (and in a lot of cases won't want to find out either), but are likely to try it just to find out.

It might be possible to make them more discoverable for those who know what to look for though.

FeepingCreature:
I mean, it's not like danbooru is any safer in that regard.

Put a link to the file in the README? I actually kinda like that - it suggests to people they can add their own tags.

Put a link to that part of the README on the settings page? Discoverability is a problem, yeah.

Acly:
danbooru can't be called safe, but you don't hit somebodys bodily fluids as the first suggestion for basically everything you type. Maybe better solution would be to use a "safe" tag database as a default, with opt-in explicit stuff, in a way people know what to expect.

I added a folder button: https://github.com/Acly/krita-ai-diffusion/commit/7ed2f8e256d3d5a7b199a9e59390265c4e7c04d9

FeepingCreature:
Actually, splitting both into "SFW" and "NSFW" files seems like the way to go, right? Of course, easier said than done...

I'd basically need a dump of the entire tag metadata for both sites to start.

Acly:
Well there's hierarchies of tag groups, I guess it's doable! Not sure how many would actually end up using them that way though.

FeepingCreature:
Danbooru has a dump at https://console.cloud.google.com/bigquery?project=danbooru1 , e621 has a dump at https://e621.net/db_export/ . Tomorrow I'm gonna take those and see if I can reconstruct the tag csv files, but this time separated in "tags appearing on sfw posts" and "tags appearing on all other posts".

Danbooru separates out "general", "sensitive", "questionable" and "explicit". How do you want the tag split?

FeepingCreature:
I guess if we're going "SFW" vs "NSFW", I should put all "sensitive" and below in the NSFW pile?

> Ecchi, sexy, suggestive, or mildly erotic. Skimpy or revealing clothes, swimsuits, underwear, images focused on the breasts or ass, and anything else potentially not safe for work.

Issue #818, opened 2024-06-16
Docker image aclysia/sd-comfyui-krita is not up to date


shahrin014:
I notice the docker image in https://hub.docker.com/r/aclysia/sd-comfyui-krita/tags is v1.16.1

The latest version is v1.18.1

I tried to automate this using github actions  but the free runner runs out of space.

Perhaps we can automate the docker build in a runner hosted on a cloud service?

moiri-gamboni:
The permanently hosted solution you're talking about is Interstice? 
In any case, personally I would appreciate even a minimal image that's up to date, since I saw there were a lot of model changes in a recent update. 
But I'm wondering if it would not make more sense to have an image without the models altogether and have them downloaded in the post-start script?

Acly:
> The permanently hosted solution you're talking about is Interstice?

Yes

I uploaded an up-to-date image now, tag is `v1.20.0` (next version) - but it should also work with v1.19.0. Only tested it local so far.

> But I'm wondering if it would not make more sense to have an image without the models altogether and have them downloaded in the post-start script?

That would be an improvement I think. It means there's few more things that can go wrong since it doesn't all come in one pre-built blob. But it would make building the image much easier and reduce duplication.

Adding a copy of this repo to the docker image and running download_models.py should work. Maybe it's possible to forward arguments to the script by editing the "Container start command" in the runpod template? That way people could pass for example `--no-sd15` and considerably decrease download size if they're not interested in SD1.5.

Acly:
`aclysia/sd-comfyui-krita:v1.22.0` is the first "barebone" image which downloads models after start-up.

https://github.com/Acly/krita-ai-diffusion/wiki/Cloud-GPU#-customization

PR #973, opened 2024-07-25
Provide a separate Docker image without checkpoints

daniel-richter:
Context: I'm using the [sd-comfyui-krita](https://hub.docker.com/r/aclysia/sd-comfyui-krita/) via Runpod. It takes ages to deploy a pod (the image is 38.81 GB) but I don't rely on any of the checkpoints it provides. 

Would it be possible to provide an additional docker image without any checkpoints? I download my own checkpoints and add them manually.

With a multistage Dockerfile it is possible to built at a certain stage and create separate tags. E.g. with the [modified Dockerfile](https://github.com/daniel-richter/krita-ai-diffusion/blob/ffc3d74e53c13f715b52d892e0227f445b118e98/scripts/docker/Dockerfile):

```sh
docker build --target base -t aclysia/sd-comfyui-krita:v1.2.3-no-checkpoints .
docker build --target full -t aclysia/sd-comfyui-krita:v.2.3 .
```


daniel-richter:
I just noticed [this discussion](https://github.com/Acly/krita-ai-diffusion/issues/818#issuecomment-2211044874):

> > But I'm wondering if it would not make more sense to have an image without the models altogether and have them downloaded in the post-start script?
> 
> That would be an improvement I think. It means there's few more things that can go wrong since it doesn't all come in one pre-built blob. But it would make building the image much easier and reduce duplication.
> 
> Adding a copy of this repo to the docker image and running download_models.py should work. Maybe it's possible to forward arguments to the script by editing the "Container start command" in the runpod template? That way people could pass for example --no-sd15 and considerably decrease download size if they're not interested in SD1.5.

Acly:
Having multiple docker images doesn't really scale well though. Now we have:
* SD1.5 with checkpoints
* SD1.5 without checkpoints
* SDXL with checkpoints
* SDXL without checkpoints
So 4 images? 4 runpod templates? I don't expect people to start composing docker images.

That's why I like the download script idea, just put the ComfyUI install with required custom nodes and dependencies into the dockerfile, and start to download models after it's up. The `download_models.py` already provides good options to select sets of models to fetch.

daniel-richter:
Here is a solution that does not copy any models into the Docker image but the `download_models.py`.

**Walkthrough**:
- Include `download_models.py` into the docker image.
  - To have the script and the dependencies available, the docker context had to be extended to the whole repo (Docker build can't access files outside of the context).
    - Use `./scripts/docker/...` for the existing `COPY` commands.
    - The command to build the image now is slightly different: 
      ```sh
      docker build -t aclysia/sd-comfyui-krita:{version} . -f scripts/docker/Dockerfile`
      ```
  - `download_models.py --minimal --no-checkpoints` will be used for the Docker image.
- Run `download_models.py` during startup of the image (in `pre_start.sh`)
- Arguments that are passed to `start.sh` will be forwarded to `pre_start.sh` & `post_start.sh` and eventually will be passed as arguments to `download_models.py`.

**Things to Discuss**
- We could omit baking the required models into the Docker image (~5.5GB with required models, ~1.6GB without models). The required models would be downloaded during container startup anyways.
- The "required_models" config already contains some SD1.5 models but no SDXL models. 
- The parts "Cleaning up cached models" and "Downloading new models" could be removed from `docker.py`, right? (The directory won't be used to build the image.)

Acly:
> To make sure the script and its dependencies are available, we had to extend the Docker context to the whole repo (Docker build can't access files outside of the context).

This is a bit problematic, Docker copies the entire "build context" to the Docker Daemon during build. Even things that don't end up used. For a fresh clone it's not a problem, but if you develop there's a lot of stuff that ends up in there (mine is 85GB). So that will take ages to transfer.

I suggest to copy `download_models.py` and `ai_diffusion` to the docker folder instead. This can be done in `docker.py`. I'm actually doing something like that in a different place, so this code should work:
```python
root_dir = Path(__file__).parent.parent.parent

def copy_ai_diffusion_modules():
    plugin_dir = root_dir / "ai_diffusion"
    target_dir = docker_dir / "scripts" / "ai_diffusion"
    target_dir.mkdir(parents=True, exist_ok=True)
    for src in plugin_dir.glob("*.py"):
        shutil.copy(src, target_dir)
    shutil.copytree(plugin_dir / "language", target_dir / "language", dirs_exist_ok=True)
    websockets_dir = target_dir / "websockets" / "src"
    websockets_dir.mkdir(parents=True, exist_ok=True)
```

Acly:
> We could skip baking the required models into the Docker image (~5.5GB with required models, ~1.6GB without models). The required models would be downloaded during container startup anyway.

Yes might as well do that.

> The parts "Cleaning up cached models" and "Downloading new models" could be removed from docker.py, right? (These models are not used to build the image.)

Yes. I'm currently reusing the models folder in some tests, but I will move it elsewhere. No point keeping it in the docker build context.

> https://github.com/Acly/krita-ai-diffusion/issues/818#issuecomment-2210811535 to download the models in the post-start script. Currently, it's done before starting the server. Is there any preference where to place the download script?

Good question. The download should definitely happen _before_ ComfyUI is started. But it might actually be better to do both in post-start, that way SSH will be up already. Do you know what happens if something goes wrong (download failure due to connection issues for example)? Is the whole start-up aborted or it just continues with the rest? The way shell scripts work I assume it's the latter. So it probably just doesn't matter much.

> Given the new setup (smaller image, larger downloads), what would be the best values for the Runpod template parameters "Container Disk (temporary)" (currently 5GB) and "Volume Disk (Persistent)" (currently 50GB)?

The way I understand it, `/` will use "Container Disk", so downloading models there might actually not work with 5GB.
But it makes more sense to download models to "Volume Disk" anyway, which is mounted to `/workspace`. That way you can stop the Pod and restart it later without having to re-download models.

So downloading models to `/workspace` and adapting the `extra_model_paths.yaml` might be a good idea.

50 GB should be enough for all models, though barely... maybe using a slightly bigger default is good.

> I added an argument to filter the checkpoints to download

I'm okay with keeping it. But if it's not compatible with `--minmal` it would be nice to put a `assert` to raise it as an error  directly.

daniel-richter:
> > We could skip baking the required models into the Docker image (~5.5GB with required models, ~1.6GB without models). The required models would be downloaded during container startup anyway.
>
> Yes might as well do that.

With all `no-*` parameters in place, only "CLIP Vision model" and "MAT Inpaint" will be downloaded. Should we add parameters to skip these two models, too?

> [...] Docker copies the entire "build context" to the Docker Daemon during build. Even things that don't end up used. For a fresh clone it's not a problem, but if you develop there's a lot of stuff that ends up in there (mine is 85GB). So that will take ages to transfer. [...] I suggest to copy download_models.py and ai_diffusion to the docker folder instead. This can be done in docker.py.

Good to know! Changed in 75c79bc6a02bc9fe52c9b55544846d9f8bbbaff0

> But it makes more sense to download models to "Volume Disk" anyway, which is mounted to `/workspace`. That way you can stop the Pod and restart it later without having to re-download models. So downloading models to `/workspace` and adapting the `extra_model_paths.yaml` might be a good idea.

Yeah, on my way home I noticed that, too. Changed in aa3573fe267ae4aa658b47d84ad4f6df737387a5. Now, the models should survive a container restart/resume. I it's okay, I've set up a directory symlink `/models` pointing to `/workspace/models` to make it more backward-compatible.

> The download should definitely happen _before_ ComfyUI is started. But it might actually be better to do both in post-start, that way SSH will be up already. Do you know what happens if something goes wrong (download failure due to connection issues for example)? Is the whole start-up aborted or it just continues with the rest? The way shell scripts work I assume it's the latter. So it probably just doesn't matter much.

- With cced200fdd7f5d2f47b81b58d91b6826b4604bb0, I should be possible to re-trigger interrupted downloads (before, an interrupted download would have been skipped because the file was already there.)
- In my opinion, the shell script would cancel the execution if any command returns an error code (by default). But that would be fixable (just append `|| true` to the download script command).
- The question is, what should we set as the default behavior in that case? If the "download_models.py" command fails, you'd have to manually re-trigger it (you'd first have to realize there was an error and then know what to do in that case).
- We could move `setup_ssh` some lines up in the `start.sh` command...

> > I added an argument to filter the checkpoints to download
>
> I'm okay with keeping it. But if it's not compatible with `--minmal` it would be nice to put a `assert` to raise it as an error directly.

The problem is/was, that the parameters are subtractive. If you use `--minimal`, there is just one checkpoint included - passing another checkpoint wouldn't have any effect. The real question is what minimal really is for. All it includes are the "required models," which is SD1.5, but not SDXL, as well as an SD1.5 checkpoint, Realistic Vision (Photography). 
Is it possible to express `--minimal` with other filter parameters?

Acly:
`--minimal` is used for some tests, also for CI on GitHub, it's minimum models required to make a successful connection.

I don't think we need an option to skip the clip-vision & MAT models, they're required and dont' have alternatives. They don't have to be part of the Docker image though (just don't run download_models.py there at all).

Looks good to me now! Automatic retry and hash checks could be nice to make downloads more robust, but I'm happy to leave it to future work.

daniel-richter:
> But if it's not compatible with --minmal it would be nice to put a assert to raise it as an error directly.

There is an assertion for the `--checkpoint` parameter now (checks, if the defined checkpoint identifier exists) (85a58a9).

> They don't have to be part of the Docker image though (just don't run download_models.py there at all).

Done (518eab0).

> Automatic retry and hash checks could be nice to make downloads more robust, but I'm happy to leave it to future work.

Automatic retry is now available (ce18782). There is a parameter `--retry-attempts` available (number of retry attempts for downloading a model, default is 5) as well as a parameter `--continue-on-error` (continue downloading models even if an error occurs, default is false but it's activated in `pre_start.sh`).

daniel-richter:
For the readme/wiki I would suggest the following text:

> ### Customizing the Container Start Command
> 
> You can customize the Docker start command to modify the default behavior by configuring the "Container Start Command" field. This allows you to override the default command with any parameters you would like to pass to `start.sh`.
> 
> To do this on RunPod.io: Before deploying your pod, click "Edit Template" to access the "Pod Template Overrides" and set the "Container Start Command" to the command you want.
>
> ### Examples
> 
> Here are some examples of container start commands for different use-cases:
> 
> 1. **Only use custom SDXL/Pony checkpoints that you will download later manually**:
>    ```
>    /start.sh --no-sd15 --no-checkpoints
>    ```
>    This command will skip SD1.5 models, download only the required SDXL models, and won't download any default checkpoints.
> 
> 
> 2. **Only use specific SD1.5 checkpoints**:
>    ```
>    /start.sh --no-sdxl --checkpoint="realistic-vision" --checkpoint="dreamshaper"
>    ```
>    This command download specific default SD1.5 checkpoints (`Realistic Vision (Photography)` and `DreamShaper (Artwork)`) and skip SDXL models.
> 
> 
> ### Parameters
> 
> Below is a list of all the parameters you can pass to `start.sh`:
> 
> - `--no-sd15`: Skip SD1.5 models.
> - `--no-sdxl`: Skip SDXL models.
> - `--no-checkpoints`: Skip default checkpoints.
> - `--checkpoint`: Specify the identifier of the default checkpoint to download. Can be used multiple times.
> - `--no-upscalers`: Skip upscale models.
> - `--no-controlnet`: Skip ControlNet models.
> - `--no-inpaint`: Skip inpaint models.
> - `--prefetch`: Fetch models which would be automatically downloaded on first use.
> - `--retry-attempts`: Number of retry attempts for downloading a model. Default is 5.
> - `--continue-on-error`: Continue downloading models even if an error occurs.
> - `-v`, `--verbose`: Print URLs and filepaths.
>

daniel-richter:
There was some confusion about how to get the version number in `download_models.py` (the explicitly declared variable wasn't used, the version itself is declared multiple times [`resources.py` and `__init__.py`]). I've updated it to use the `resource.version` (e9df3fefd7a7e4e089c1dd25cec2bd7f8b591d1a).

As a bonus, we were able to cut down on the number of dependencies for `download_models.py`, so now we just need to copy `scripts/download_models.py` and `ai_diffusion/resources.py` into the container (14a08b214a593ca45aacb909de297422889abf76).

I also added a dry run of `download_models.py` during the image build to create the model directories (202496bc937ca241f2f48e8a9ef61c3255a86b03).

Acly:
> Would it make more sense to call the parameter no-preprocessors?

I do want it to be all prefetch models. It would make sense to make it use the `resources.prefetch_models` list though. It just happens to be equivalent with `ResourceKind.preprocessor` at the moment.

> I've updated it to use the `resource.version`

I think that's fine, the docker images mostly track the resource version. Plugin and resource versions do not always match, there's a reason they're separate. The resource version only changes when any of the resources do, which is not always the case for new plugin release.

Acly:
Thanks for working on this!

I might change the arguments of `download_models.py` to be additive instead of subtractive before building & releasing. I think it makes more sense in the long run, with more and more models being added. Case in point, there is Flux now with another 17G, not sure it will stay but I wouldn't want it to download by default.

moiri-gamboni:
I have some old code that compares content-length on existing files. I can't really work on a PR at the moment, but feel free to copy it if you find it helpful. 
https://github.com/moiri-gamboni/krita-ai-diffusion/commit/01b3385d419ef7c9ba740e3595cb85f4d230dea8

Acly:
I think downloads that don't complete are covered now via temporary file during download, so the content length maybe doesn't have as much use anymore.

I made some more docker changes, updates examples and list of arguments: https://github.com/Acly/krita-ai-diffusion/wiki/Cloud-GPU#-customization

It follows more the scheme used by the UI installer now. Use `/start.sh --sdxl` to get all the basic requirements for SDXL but no checkpoints or anything optional.

PR #988, opened 2024-07-27
Brazilian Portuguese Translation

flaviohr:
- added pt-br Translation
- fixed some spell erros in Spanish translations, and reverted "Cargas de trabajo" back to "Workload"


RaHiMu:
I corrected some expressions that were a little confusing in Spanish or that needed more precision

PR #1043, opened 2024-08-07
Split Danbooru tags into SFW and NSFW, re-add e621.

FeepingCreature:
Follow-up to https://github.com/Acly/krita-ai-diffusion/pull/953

Instead of just copypasting them from the previous PR, I added a README that documents the steps taken to produce these files.

Broadly, "SFW" tags are tags that appear on "general" (danbooru) or "safe" (e621) posts, and "NSFW" tags are all others.

There's a bit of imprecision; for instance, e621 has `object_in_ass` appearing (correctly!) on safe (!) pictures... three times vs 30k in NSFW images :-) So the e621 tag sorter defines tags that are >99% on NSFW images as NSFW; this seems to be working.

And of course, massive shout-outs to Claude 3.5 Sonnet, without whom all of this would have been much slower, and who wrote ~all the code in this PR.

(Why did I add column names as the first line? That's how BigQuery exports CSVs... and I figure it's a good idea anyway so might as well.)

edit: Potential problem I just spotted: this PR doubles the amount of tags from Danbooru alone, because now `Danbooru` is 100k lines and `Danbooru NSFW` is another 100k lines. Should I drop the limit to 50k? ai_diffusion/tags is still only 8.8MB.


Acly:
>  Should I drop the limit to 50k? ai_diffusion/tags is still only 8.8MB.

How about using minimum usage as cut-off? Drop everything below eg. count 50. It doesn't really make sense for nsfw to be as large when it has way less total use.

FeepingCreature:
I mean, I just tried it and AutismMix's Clip has tag understanding of Danbooru tags with as little as 30 count, though it increasingly falls back on its common-sense understanding. For instance, cross-laced bra at 30 still works. On the other hand, characters, settings and styles at that level are hopeless, which makes sense. So my current setting is cut off characters, artists and settings at 50, but let concepts go all the way down to 20 - they're a lot thinner on the ground anyways. Does that work?

FeepingCreature:
Limited to 20 general, 50 everything else:
- Danbooru 81k
- Danbooru NSFW 7k
- e621 63k
- e621 NSFW 10k

Seems better!

Acly:
That looks reasonable. While I wouldn't exactly recommend the "SFW" versions for use at work, I think all the "offensive" (to a significant number of people) stuff is gone.

PR #1050, opened 2024-08-09
Update ru.json (09.08.2024)

Drakosha405:
updated to the latest version


Acly:
Updated the files, thanks. Additional UI for upcoming release is also mostly done, new strings are added at the end.

PR #1100, opened 2024-08-22
Add Korean language interface

sejoung:
Hello, the Korean language file has been added


Acly:
Thanks! I added the language file.

Additional UI for upcoming release is also mostly done, new strings will be added at the end of the file.

PR #1150, opened 2024-09-05
Layer Diffusion integration

paprik:
Basic implementation of ComfyUI Layer Diffusion: https://github.com/huchenlei/ComfyUI-layerdiffuse
Uses two new custom nodes and adds a new transparency checkbox for the Generate action.

Showcase:
transparency_test

### DIFF
--- a/ai_diffusion/api.py
+++ b/ai_diffusion/api.py
@@ -157,6 +157,7 @@ class WorkflowInput:
     control_mode: ControlMode = ControlMode.reference
     batch_count: int = 1
     nsfw_filter: float = 0.0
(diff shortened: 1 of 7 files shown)


paprik:
Created as draft since I likely missed something. Generate (with upscaling too) works great. Could not manage to make it work with Refine/Inpaint yet, it keeps adding a white background for some reason.

paprik:
You're right, 2-pass upscale doesn't work, I've just generated a 2k in one step. Hopefully it gets added to ComfyUI nodes soon.

I still have to try Attention injection, because the Conv one changes the model a quite a lot.

PR #1156, opened 2024-09-06
Fix issue 1137: Display ComfyUI Queue State

FeepingCreature:
Query ComfyUI queue state and use it to display count of jobs in the queue before ours.

Also use it to avoid clearing jobs that aren't ours.


FeepingCreature:
Re why didn't I fix cancel current... so far as I can tell, this can only cancel another user's task if you do it right as a task starts and ends, and that's p much unavoidable due to comfy's anemic api, ie. interrupt not having a way to specify the job id we expect.

ai_diffusion/comfy_client.py
```diff
@@ -340,9 +342,34 @@ async def clear_queue(self):
...
+
         self._jobs.clear()
 
+    async def _poll_server_queue(self):
+        queue = await self._get("api/queue")
+        server_jobs = queue["queue_running"] + queue["queue_pending"]
+        # why are they unsorted to start with...?
+        server_jobs = sorted(server_jobs, key=itemgetter(0))
+        server_jobs = [entry[1] for entry in server_jobs]
+        if not (self._jobs or self._active):
```
Acly, line 356: Not really sure about reporting general server state/activity when we're not waiting for a job. I think it's more likely to be confusing than helpful if you don't know how to inerpret the icon exactly.

ai_diffusion/comfy_client.py
```diff
@@ -340,9 +342,34 @@ async def clear_queue(self):
...
+        else:
+            first_remote_id = await self._jobs[0].get_remote_id()
+        # if we got it from _jobs or _active, this field must have been set (in _run_job).
+        first_remote_id = util.ensure(first_remote_id)
+        try:
+            offset = server_jobs.index(first_remote_id)
+        except ValueError:
+            # probably just haven't gotten the notification yet
+            return
+        await self._report(ClientEvent.foreign_jobs, "", foreign_jobs=offset)
```
Acly, line 371: Does it make sense to report the ID here? The UI would consider the job to be executing while waiting for its turn in the queue, which I think is how it already worked for the cloud client

ai_diffusion/client.py
```diff
@@ -25,6 +25,7 @@ class ClientEvent(Enum):
     disconnected = 5
     queued = 6
     upload = 7
+    foreign_jobs = 8
```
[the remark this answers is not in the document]
FeepingCreature: Oh yeah you're right, looking at the cloud client. So how does that work, we indicate queued for the first job in the list?

ai_diffusion/client.py
```diff
@@ -34,6 +35,7 @@ class ClientMessage(NamedTuple):
     images: ImageCollection | None = None
     result: dict | None = None
     error: str | None = None
+    foreign_jobs: int | None = None
```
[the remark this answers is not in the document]
FeepingCreature: Yeah will go for it as I fold this into ClientEvent.queued.

ai_diffusion/comfy_client.py
FeepingCreature: I'll drop it in the next version. I don't really like the whole iconography, honestly.

ai_diffusion/comfy_client.py
Acly: Yea the cancel is a bit ambiguous in this case. But from user perspective I think it makes sense to track the active job (the earliest submitted by the user and not finished yet), and consider all follow-up jobs enqueued. The active job may be in a "waiting" state while the server is busy.

"Cancel active" should probably remove the active job from the server queue if it isn't processing yet in that case.

ai_diffusion/client.py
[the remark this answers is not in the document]
FeepingCreature: done

ai_diffusion/client.py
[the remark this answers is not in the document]
FeepingCreature: done

ai_diffusion/comfy_client.py
FeepingCreature: dropped

ai_diffusion/comfy_client.py
FeepingCreature: I'm gonna leave that out for now cause I think you underestimate how ugly it's going to be. Right now the job queue tracks the server-side job queue pretty closely, which is why we can just call `interrupt()` to clear out the running job. But that's state that's distributed over like three classes, ie. client, jobs and the comfy server. If we start pretending that a queued job is an active one, even in the job queue, then the clients have to keep their own account of what's going on and redirect interrupt() to *sometimes* do `/interrupt` and sometimes `/api/queue: delete`.

ai_diffusion/model.py
```diff
@@ -408,9 +410,13 @@ def handle_message(self, message: ClientMessage):
             return
 
         if message.event is ClientEvent.queued:
-            self.jobs.notify_started(job)
-            self.progress = -1
-            self.progress_changed.emit(-1)
+            if message.queue_length is not None:
+                self.queue_length = message.queue_length
+                self.queue_length_changed.emit(message.queue_length)
+            if message.queue_length is None or message.queue_length == 0:
```
FeepingCreature, line 416: This is where the "don't start the job until it's actually running" thing moved to.

FeepingCreature:
Something's wrong with the code as it stands, queueing pictures makes the progress bar flicker. I'll check the logs tomorrow.

ai_diffusion/model.py
Acly: I'd suggest to consider it started in all cases, like before (just add the queue length if available).

ai_diffusion/comfy_client.py
Acly: Yes I thought the interrupt call might just become:
```
if self._active_job:
  self._post("/interrupt")
else:
  self._post("/api/queue", {"delete": self._jobs[0].remote_id})
```
(something like that)

But maybe not?

FeepingCreature: Idk, if anything, shouldn't it be done in model, ie. `cancel`? I just want to minimize the amount of state that is lying.

ai_diffusion/model.py
FeepingCreature: I guess I don't see what it buys you. Why should we consider job started if the job is not started?

ai_diffusion/comfy_client.py
Acly: It's not lying, only a different perspective.

The Model reflects what the user is interested in. The UI presents it. The Client maps the inner workings of whatever server it's written for to the outside model for easy consumption. Knowledge about inner workings of the server shouldn't leave the Client.

UI absolutely wants to paper over details users aren't interested in. I disagree on "close to the surface", complexity should be dealt with at the lowest layer possible, rather than leaking to upper layers, so they don't have to bother with it.

As a user I hit Generate, and from that moment I've started a job to create an image. It may involve several stages (waiting for spot in queue, uploading, generating, downloading) and it's done when I can see the image.
* I can enqueue further jobs, but those won't start until the current job is done. 
* I can cancel the current one if I'm no longer interested in that particular image, and it should interrupt processing at whatever stage it is in (I don't care).
* I can cancel enqueued jobs, and it should remove them from the queue without affecting the current job.

The client is responsible for communication with the server, so it needs some more detailed representation of a job. It may not consider a job active unless it's actually executing, so it knows what to do when it's supposed to be canceled (dequeue vs interrupt).

I think that's the desired behavior at least. Maybe the naming in the code could reflect it better?

FeepingCreature: I think I've figured out why this bewilders me so. In that case, isn't `job.any_executing` silly? We shouldn't care if a job is "executing" at all, and the cancel button should never be disabled unless there's *no* jobs in the queue. `any_executing` should be removed, and `JobState` should just be `queued, finished, cancelled`.

I think that's a valid refactor, and I'll do it if you think it's a blocker, but at any rate it does not belong in this PR.

FeepingCreature: I think that, if the "cancel first" button is always shown, and always removes the first element, there genuinely isn't any code left that conditions on `JobState.executing`. Am I missing any?

Acly: You mean any(queued) and any(executing) are equivalent? Mostly I guess. If you start a job and it errors early (connection failure or server enqueue failing directly) it never goes into executing.

FeepingCreature: I mean, does it matter at that point? With that refactor, what is actually still dependent on executing?

Acly: I'm sure making executing implicit (oldest non-finished-non-cancelled job) works too, but I kinda like that it relies on client feedback and is an explicit state. I don't see a strong argument one way or the other, but if there's no strong reason to change, keep what already works.

FeepingCreature: I'm just hopelessly confused now. If you like that job state relies on client feedback, why do you want me to mark the first job, that the client has not started executing yet, as executing?

Acly: Because I consider standing in line as part of executing my job! You know, pushing a little, trying to advance a spot or two when those in front are looking the other way. Things a good client does.

Call it "active" or "in_progress" or something if executing is so confusing. The important part is that the job is the focus of attention and something is being done about it. That's the job I care about, and which the UI progress bars and indicators refer to. Surely that's worth a special state? (compared to all the other jobs that are either old, finished and forgotten or distant future)

FeepingCreature:
Ah ftr I'm not gonna touch this until end of the week cause I'm on a conference :)

PR #1177, opened 2024-09-13
Downloader: use truststore for system cert trust when available

Rotonen:
Now with the proxy support, one can easily hit resigning proxies with the model downloader, and naturally `certifi` does not come with support or a way to inject support for arbitrary PKI solutions.

Truststore is an elegant way around this problem, it lets one trust the system certificate store instead of certifi: https://pypi.org/project/truststore/

I find the "just capture all things TLS" pattern the neatest if truststore is installed, but I'm open for debate on the implementation details.

https://truststore.readthedocs.io/en/latest/index.html#using-truststore-with-aiohttp

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,6 +7,7 @@ requests
 
 # Testing
 aiohttp
+truststore
 python-dotenv
 markdown
 numpy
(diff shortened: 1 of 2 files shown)


Acly:
Thanks! truststore seems like it would have solved a lot of headaches if it was the default...

Rotonen:
Yes, the whole http-in-python ecosystem has a long and varied history of stubbornness around the topic of trusting arbitrary PKI.

And sorry, was not aware of the formatting rules on the repo. Should I put forth a further PR to flip that bulid green?

Acly:
I added the new-lines after merge, so all good.

Issue #1102, opened 2024-08-23
SSL Error on MacOS still


qr3325:
This seems to occur despite the fix in #840 / #841.

Client log: 
```
2024-08-23 01:23:57,585 ERROR Error setting SSL_CERT_FILE on MacOS: No module named 'certifi'
2024-08-23 01:23:57,595 INFO Loaded 11 sampler presets from /Users/qr3325/Library/Application Support/krita/pykrita/ai_diffusion/presets/samplers.json
2024-08-23 01:23:57,638 INFO Extension initialized, Version: 1.22.0, Python: 3.10.7 (main, Apr 17 2024, 20:18:29) [Clang 15.0.0 (clang-1500.3.9.4)]
2024-08-23 01:23:58,684 INFO Connecting to http://127.0.0.1:8188
2024-08-23 01:24:43,352 INFO Connecting to https://fnqdspfyb1aykl-3001.proxy.runpod.net
2024-08-23 01:24:44,342 ERROR Error: Could not establish websocket connection at wss://fnqdspfyb1aykl-3001.proxy.runpod.net: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:997)
Traceback (most recent call last):
  File "/Users/qr3325/Library/Application Support/krita/pykrita/ai_diffusion/comfy_client.py", line 94, in connect
    async with websockets_client.connect(f"{wsurl}/ws?clientId={client._id}"):
  File "/Users/qr3325/Library/Application Support/krita/pykrita/ai_diffusion/websockets/src/websockets/legacy/client.py", line 629, in __aenter__
    return await self
  File "/Users/qr3325/Library/Application Support/krita/pykrita/ai_diffusion/websockets/src/websockets/legacy/client.py", line 647, in __await_impl_timeout__
    return await self.__await_impl__()
  File "/Users/qr3325/Library/Application Support/krita/pykrita/ai_diffusion/websockets/src/websockets/legacy/client.py", line 651, in __await_impl__
    _transport, _protocol = await self._create_connection()
  File "/Applications/krita.app/Contents/Frameworks/Python.framework/Versions/Current/lib/python3.10/asyncio/base_events.py", line 1089, in create_connection
    transport, protocol = await self._create_connection_transport(
  File "/Applications/krita.app/Contents/Frameworks/Python.framework/Versions/Current/lib/python3.10/asyncio/base_events.py", line 1119, in _create_connection_transport
[...]

PR #1181, opened 2024-09-14
Fix CERTIFICATE_VERIFY_FAILED errors on macOS

voronin-ilya:
Fixes https://github.com/Acly/krita-ai-diffusion/issues/1102

### DIFF
--- a/ai_diffusion/comfy_client.py
+++ b/ai_diffusion/comfy_client.py
@@ -28,7 +28,8 @@ from . import resources, util
 
 if util.is_macos:
     import os
-    os.environ["SSL_CERT_FILE"] = "/etc/ssl/cert.pem"
+    if "SSL_CERT_FILE" not in os.environ:
+        os.environ["SSL_CERT_FILE"] = "/etc/ssl/cert.pem"
 
 
 @dataclass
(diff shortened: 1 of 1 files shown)


Acly:
This feels somehow even more brittle than the previous solution. At least I couldn't (quickly) find any information that confirms the cert file will always exist in this location (depending on macOS version and what packages are installed)...?

If the issue is that the environment variable typically isn't set at all, maybe it would be better to only set it if it doesn't exist. That way it would still be possible to set a custom value outside at least, without it being overwritten.

voronin-ilya:
I am unable to confirm that this file is present on every macOS version (although some Googling suggests it most likely is). Which solution do you prefer: setting SSL_CERT_FILE only if it’s not already set, or adding certifi as a submodule and including it as done with the websockets library? The current solution isn’t really working since certifi is not bundled with the Pykrita environment.

Acly:
I think adding certifi as submodule is overkill.

What _might_ work is calling `python -m certifi` as a subprocess, so it uses system python instead of Krita's embedded. But this would require python to be installed on the system, which is probably not always the case either.

I generally also don't mind the fixed path, I just think the env shouldn't be overwritten if it was set already, to still allow it being set outside if the hard-coded path doesn't work. Or do that as a fallback.

PR #1236, opened 2024-09-30
Make positive input prompt drag resizable.

FeepingCreature:
As per mention in Discord.

Problem: Drag bar currently occupies five invisible pixels at the bottom of the text widget. Not sure how to fix that in the layout.

### DIFF
--- a/ai_diffusion/ui/region.py
+++ b/ai_diffusion/ui/region.py
@@ -1,7 +1,16 @@
 from __future__ import annotations
 from enum import Enum
 from PyQt5.QtWidgets import QWidget, QLabel, QToolButton, QHBoxLayout, QVBoxLayout, QFrame, QMenu
-from PyQt5.QtGui import QGuiApplication, QMouseEvent, QResizeEvent, QPixmap, QImage, QPainter, QIcon
+from PyQt5.QtGui import (
+    QGuiApplication,
+    QMouseEvent,
+    QResizeEvent,
+    QPixmap,
+    QImage,
+    QPainter,
+    QIcon,
+    QFontMetrics,
+)
 from PyQt5.QtCore import QObject, QEvent, Qt, QMetaObject, QSize, pyqtSignal
 
 from ..root import root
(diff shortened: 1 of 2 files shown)


Acly:
It works pretty well.

Would be nice to have a visual indicator, and to avoid the extra layout space. You can put the widget on top of the text box by not adding it to the layout and using `move` to set a fixed position instead.

This is how the translation button is positioned, same code should work (could also consider moving it to the same place): https://github.com/Acly/krita-ai-diffusion/blob/main/ai_diffusion/ui/region.py#L393

As for visual indicator, something like this?
image

Diagonal lines in the bottom right are also typical, but the translation button is already there and moving it looks a bit weird:
image

krigeta:
any planning to add this feature with the main branch?

FeepingCreature:
Sorry! I'll try to get to it.

krigeta:
> Sorry! I'll try to get to it.

np, please let me know when you are done

FeepingCreature:
> As for visual indicator, something like this?

How bout now?

Acly:
> How bout now?

Looks lovely, but only when negative prompt is enabled (no button otherwise) and you're using a light theme (barely visible otherwise).

> I don't understand why the test is red tbh.

Don't worry about it, is fixed but still on a branch

FeepingCreature:
- [ ] fix in light mode
- [ ] off-by-one pixel error
- [ ] fix with negative prompt off

Acly:
Nice!

 found one more thing, you can get it to this state:
image

But I think I know why: when you toggle neg prompt in settings, it uses `isVisible` to determine if it's already enabled on the positive prompt. This doesn't work properly for the text box in other then the active workspace (eg. Live box when you are in Generate mode), because it inherits visibility from parent.

I can just fix it after merge though since I was curious and more or less did it already now :)

FeepingCreature:
Ah yeah oops - I ran into that a bunch, but I didn't think to check for switching workspaces.

Cheers!

Issue #628, opened 2024-04-18
Update live preview only after a deliberate change rather than mid-change


lboklin:
Currently the live preview will generate new images constantly while I am in the process of making changes to the image or prompt, even mid-stroke/mid-typing. This is unnecessary and will only delay the final generation based on the change I mean to make, not to mention the constant output of images that will inevitably be replaced (it can be frustrating to see incidentally desirable results and not being able to apply or reproduce it).

For this, some basic heuristics would be useful, if possible, to hold off on generating previews until there is a pause in prompt editing or a stroke has been completed. Krita itself decides which actions should be undo-able, and that heuristic could be a good starting point at least as it pertains to image modifications.

Acly:
So the idea of live preview is that it's (more or less) "live", ie. it updates in less than a second. I know this is not achievable on a lot of hardware, but there is the "classic" generation mode for handling it as longer running background task. What you're suggesting make me thing that it's more suited for what you want...

To clarify some points:
1. Live requests are never queued, a new one is only started when the previous one has finished. Work queued previous from other modes is completed first.
2. Cancelling live requests does not make sense, the idea is that it updates while you draw/type/work. If it cancelled on every change, it would never generate an image at all in that case - not until you sit back and actively wait without touching anything. It would be very frustrating.
3. Only updating for "deliberate changes" makes some sense, but is difficult to implement. Python plugins don't have access to Krita's undo stack, or even any events that indicate things have changed. It's currently implemented via periodic polling and full comparisons of the pixel data... not much to go on. For text changes it would be under control, but even that seems a lot of work for minor effect.

And maybe as a workaround: You can set a hotkey for "Generate" in Krita's keyboard settings. It also works in live mode when it's not active. So if you prefer the preview on the side, you can use it to manually trigger generation with 1 button press.

lboklin:
What you say makes sense and I agree that neither cancelling nor queueing make much sense. I would however consider (3) a significant improvement to the UX. Without access to the undo stack, is there perhaps a way to query activity in some other way to at least implement some sort of delay for when modifications are being actively made?

> And maybe as a workaround: You can set a hotkey for "Generate" in Krita's keyboard settings. It also works in live mode when it's not active. So if you prefer the preview on the side, you can use it to manually trigger generation with 1 button press.

That's a good workaround, thanks for the tip!

PR #1248, opened 2024-10-08
Add a setting to cancel live jobs upon change

modelflat:
Hi, thanks for this amazing plugin!

I've noticed an inconvenience with the way the jobs are scheduled in the live mode, and thought to contribute a fix I implemented for myself.

Current implementation of live mode will try to schedule jobs "sequentially":

- the first change causes a job to be scheduled
- we stop listening for changes until the job is finished
- if the job wasn't for the latest change, a new job is scheduled, etc.

This often results in having to wait for the intermediate results to complete, and is not always convenient. 

This change adds a setting, which, if toggled on, instructs Live mode to clear the queue and interrupt any running jobs upon receiving a new change, thus avoiding having to wait for intermediate results.

### DIFF
--- a/ai_diffusion/comfy_client.py
+++ b/ai_diffusion/comfy_client.py
@@ -301,7 +301,11 @@ class ComfyClient(Client):
                             ClientEvent.progress, self._active.local_id, progress.value
                         )
                     else:
-                        log.error(f"Received message {msg} but there is no active job")
+                        if settings.live_cancel_jobs_on_change:
+                            # likely to be an out-of-order progress from cancelled/interrupted job so don't log these
+                            pass
+                        else:
+                            log.error(f"Received message {msg} but there is no active job")
 
                 if msg["type"] == "executed":
                     job = self._get_active_job(msg["data"]["prompt_id"])
--- a/ai_diffusion/model.py
+++ b/ai_diffusion/model.py
@@ -348,6 +348,8 @@ class Model(QObject, ObservableProperties):
         if input != last_input:
             self.clear_error()
             params = JobParams(bounds, conditioning.positive, regions=job_regions)
+            if settings.live_cancel_jobs_on_change:
+                await self._cancel_everything()
             await self.enqueue_jobs(input, JobKind.live_preview, params)
             return input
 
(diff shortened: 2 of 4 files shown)


Acly:
There is some discussion about this in #628 

The gist is that I'm not convinced cancel is a good idea, mainly because it means you have to actively restrain yourself from making any kind of change to avoid accidentally interrupting image generation when you don't want to. The basic idea of Live mode is that your are making small changes constantly and get (reasonably fast) feedback.

If you find yourself waiting for results before continuing to work, the queued workflow or manual trigger via hotkey  might be a better fit?

I also have some issues with the implementation, usually there is a flow like this:
* cancel active job
* generation is interrupted on the server
* server send message that job has been cancelled
* client receives message and forwards to model / jobs
* job state is set to cancelled (and may be removed from list)
* live mode can continue by scheduling the next image as a result of previous job being completed/cancelled

 `_cancel_everything` is taking a kind of shortcut, and maybe getting away with it because all jobs should be nuked afterwards on all sides... but  it shouldn't be necessary

modelflat:
@Acly Thanks for the feedback!

> it means you have to actively restrain yourself from making any kind of change to avoid accidentally interrupting image generation when you don't want to
After having worked with the plugin some more and getting used to it, I realize that this is a very important point. Even if the live generation takes a while (up to 4 seconds on my setup), it is still more meaningful for it to run sort of "in the background" while you are working uninterrupted. 

Regarding the suggestion to use a hotkey to generate: while I believe this might be a good workaround for _very_ slow setups, I still consider it a very significant UX hit as these are additional actions you have to actively take, as opposed to simply keeping an eye on the preview panel.

So, there are two potential improvements I see for the live generation process still:
1. Debounce the change events (as already suggested in #628). This should be of help at least in the case of rapid modification and/or on weaker setups. I believe buffering the events by 100-200ms (or configurable) window should reduce unwanted generations, while preserving the desired properties of live updates
2. Have a backlog of the last N images generated in live mode. Again, this applies more to slower setups like mine. At the sketching stage, the model can sometimes produce interesting results which you might wanna build from, but it's hard to go back to it if you've been making changes rapidly

modelflat:
Continued in #1412

PR #1314, opened 2024-10-23
Update ko.json to v1.26.0

sejoung:
Update ko.json to v1.26.0


sejoung:
It seems that lint errors are appearing in files I didn’t modify.

Acly:
> It seems that lint errors are appearing in files I didn’t modify.

Yes those came from main, fixed now.

Thanks for the update!

PR #1416, opened 2024-11-18
 Positive prompts in custom workflows should honor the prompt_line_count setting 

### LINKED ISSUE #1415: Positive prompts in custom workflows should honor the prompt_line_count setting
The prompt input area tends to be too small and there's already a setting for exactly this. I can't see any downside apart from super special workflows with multiple positive prompts, but that's probably an edge case.

Pull request coming shortly. I've also added the settings button to the "Graph" workspace since it's the only one that doesn't have it and there's enough settings that apply to custom workflows, especially with this change.

mmuellerdm:
Fixes  #1415

Sorry for the revert commit, I've had some trouble with branches.

### DIFF
--- a/ai_diffusion/ui/custom_workflow.py
+++ b/ai_diffusion/ui/custom_workflow.py
@@ -247,8 +247,16 @@ class PromptParamWidget(TextPromptWidget):
     value_changed = pyqtSignal()
 
     def __init__(self, param: CustomParam, parent: QWidget | None = None):
-        line_count = settings.prompt_line_count if param.kind is ParamKind.prompt_positive else TextPromptWidget._line_count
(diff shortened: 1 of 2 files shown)


Acly:
> Sorry for the revert commit, I've had some trouble with branches.

You can rebase/squash them if you want, but Github can also do it when merging so no worries.

You might want to run `black` for auto-format (or set it up in your IDE) to avoid the tests failing. Looks good otherwise.

mmuellerdm:
It looked to me like the lack of type check on update was an obvious oversight, but it turns out it's actually not necessary? I have no idea why, to be honest.  I suppose it can't hurt.

PR #1421, opened 2024-11-19
Info to clipboard

### LINKED ISSUE #1366: "Copy" context menu options don't do anything in "Graph" mode.
It seems that the plugin just doesn't know where to put the data it obviously has (it's all in the mouseover popup), since it works fine in "Generate" mode. This isn't a huge issue if you know it, but somewhat of a trap since I just lost a few prompts that I thought I had saved that way. Maybe copy the data to the clipboard if it can't be properly put back into the UI? Or at least gray out the options that can't work like it already does for "Copy Style"?

The "proper" way would be to change the context menu based on the parameter nodes in the workflow, but that is obviously not trivial and has edge cases where the workflow can change along the way etc.

mmuellerdm:
New PR with own branch, fixes #1366

### DIFF
--- a/ai_diffusion/ui/generation.py
+++ b/ai_diffusion/ui/generation.py
@@ -159,11 +159,15 @@ class HistoryWidget(QListWidget):
         if params.strength != 1.0:
             title = f"{title} @ {params.strength*100:.0f}%"
         style = Styles.list().find(params.style)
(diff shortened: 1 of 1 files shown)


Acly:
Thanks for the PR!

PR #1429, opened 2024-11-21
Copy prompt parameters from custom workflows

ronan36880:
This change makes it possible to also copy the parameters used in custom workflows. It works by going through all the metadata in the job and sets the matching widget's value. 

This makes it so "Copy prompt" will now also work on custom workflows and automatically sets the values on the UI.

I had to implement a signal, but I'm unsure if it's the best approach to doing it, would love some feedback on this if possible.

### DIFF
--- a/ai_diffusion/custom_workflow.py
+++ b/ai_diffusion/custom_workflow.py
@@ -565,13 +565,24 @@ class CustomWorkspace(QObject, ObservableProperties):
         assert self._last_result and self._last_job, "No live result available"
         return self._last_result, self._last_job
 
+    def try_set_params(self, params: dict):
+        self.params = _coerce_with_fallback(params, self.params, self.metadata)
+        self.graph_changed.emit()
+
 
 def _coerce(params: dict[str, Any], types: list[CustomParam]):
-    def use(value, default):
-        if default is None:
-            return value
-        if value is None or not base_type_match(value, default):
-            return default
-        return value
+    return _coerce_with_fallback(params, {}, types)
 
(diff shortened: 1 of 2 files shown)


Acly:
It might be simpler, can you try:
```
# custom_workflow.py
class CustomWorkspace:
    def try_set_params(params: dict):
        self.params = _coerce(params, self.metadata)

# ui/generation.py in _copy_prompt
self._model.custom.try_set_params(job.params.metadata)
```

Should not require the signal, or any changes in the custom param UI part. And it will also take care of cases like having a different workflow which uses a param with the same name, but it has a different type (eg. string instead of number).

ronan36880:
I tried going with that approach, I still had to emit the `graph_changed` signal for the UI widgets to update. Should the extra coerce logic be in a separate method like it is right now, or should they both be merged into one somehow?

Acly:
> Should the extra coerce logic be in a separate method like it is right now, or should they both be merged into one somehow?

Isn't `coerce` the same as `coerce_with_fallback` and passing a empty dict `{}` as fallback?

ronan36880:
> > Should the extra coerce logic be in a separate method like it is right now, or should they both be merged into one somehow?
> 
> Isn't `coerce` the same as `coerce_with_fallback` and passing a empty dict `{}` as fallback?

Great catch! I went ahead and made the original `_coerce` call `_coerce_with_fallback` with an empty dict.

Acly:
Merged, thanks for the PR!

PR #1539, opened 2025-01-13
Improve compatibility with custom tag lists

BetaDoggo:
This is a simple change which makes the tag list loader a bit more compatible. 

Specifically, it doesn't assume that a header row exists, and will instead skip rows that lack a numeric tag type. Also, to improve compatibility with non-danbooru lists the tag category now defaults to general if the TagType doesn't match the known categories. This change is aimed specifically at e621 which has additional categories when pulling from the api which are not in the danbooru set.

My primary goal was to make my [own tag set](https://github.com/BetaDoggo/danbooru-tag-list/releases/tag/tags) work with the plugin without additional modification.


Acly:
Looks good, thanks!

PR #1568, opened 2025-01-27
Added a switch to the LoRAs in the style dialoge.

moonwhaler:
Added a switch to the LoRA list in the styles / settings dialog to activate/decativate single LoRAs. This makes experiments with different LoRAs or LoRA combinations more streamlined and intuitive.

### DIFF
--- a/ai_diffusion/api.py
+++ b/ai_diffusion/api.py
@@ -45,11 +45,12 @@ class ImageInput:
 class LoraInput:
     name: str
     strength: float
+    enabled: bool = True
     storage_id: str = ""  # Base64-encoded SHA256 hash
 
     @staticmethod
     def from_dict(data: dict[str, Any]):
-        return LoraInput(data["name"], data["strength"])
+        return LoraInput(data["name"], data["strength"], data.get("enabled", True))
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


moonwhaler:
Not sure what the pipline error is exactly... 
Screenshot_20250127_220114

Acly:
I believe the change in `api.py` should not be needed, since disabled Lora aren't added to the workflow input in the first place (filtered out in style). Looks good otherwise

> Not sure what the pipline error is exactly...

looks like this for me:
image

basically just run `ruff format` and it fix the formatting

moonwhaler:
Just did that. Hope it worked.

Acly:
Works.

Please try to remove api.py changes also:

> I believe the change in api.py should not be needed, since disabled Lora aren't added to the workflow input in the first place (filtered out in style).

moonwhaler:
Just reverted the changes to api.py - and tested this. You're right. I didn't need the change in that file.

LaVADraGoN:
Its a nice touch, jet even if disabled and confirming this, they still show in the image internal generation info/Metadata. Just like the Lora has been actually used in the generation. I think this is not intended.

PR #1593, opened 2025-02-07
Random tags and artists from e621 and danbooru

coffeegrind123:
It's useful


Acly:
The PR contains a lot of things that can't be merged like that (requiring dependencies installed into Krita's embedded python, windows only, icon downloads...). I'm also not convinced there's space for it in the UI.

What is the plan or intention here?

I can see random tags/artist maybe using placeholders in the text prompt (eg. writing `<random:artist>` would be replaced accordingly)

rktvr:
why would a random tag selection need beautifulsoup? wtf are those dependencies

PR #1651, opened 2025-03-11
Catalan translation

lliurexia:
Catalan translation

(Spain regional language)


lliurexia:
Is the check process ok? It is failing.

Acly:
> Is the check process ok? It is failing.

There was some issue with a cache I think, nothing to worry about.

Thanks for the translations!

PR #1694, opened 2025-03-31
Add Option to Evaluate Dynamic Prompts

nolan778:
### Feature: Dynamic Prompts

- Add support for dynamic prompt evaluation, enabled/disabled through a new option on the Interface UI settings page (disabled by default).
- This feature may not exactly fit with the main use case of this plugin (live editing with AI), so I have disabled it by default, but I find myself using this plugin primarily for image generation, due to its ease of user over comfyUI.  Having a robust way to easily generate varied prompts within Krita was important to me. 
- Currently only supports random sampling on variants (in-line), wildcards (.txt files), and variables, with support for various nesting combinations of all three.

### Additions

- New UI Interface setting to enable/disable (off by default)
- When enabled, both positive and negative prompts (after being merged with style prompt text) are individually parsed for dynamic prompt template syntax (variants, variables, wildcards).
- If any template syntax is found, it is evaluated by new functions in the text module and the template syntax is replaced by the intended or randomly chosen values.
- See [sd-dynamic-prompts Basic Usage](https://github.com/adieyal/sd-dynamic-prompts/tree/main?tab=readme-ov-file#basic-usage) and [sd-dynamic-prompts Template Syntax](https://github.com/adieyal/sd-dynamic-prompts/tree/main?tab=readme-ov-file#template-syntax) for information on proper syntax and capabilities.  I followed this example as closely as possible, including the glob file pattern matching notation.
- Wildcard directories are:
  - util.plugin_dir / "wildcards" (lowest priority, missing by default, useful if you want to ship any wildcards with the plugin)
  - util.user_data_dir / "wildcards" (higher priority, missing by default, user would create and add files or directory structures of files)
  - custom directories (highest priority, not currently used but could be populated via a future UI element, currently just used for unit testing)
- The dynamic prompt text evaluation functions are thoroughly unit-tested.  Wildcard files used for testing are auto-generated at unit test runtime into tests/data/wildcards and that directory is cleaned up automatically at test completion.

### Limitations:

- Job Info on the images in the history does not contain the actual final evaluated prompt that was sent to comfyUI in the CLIPTextEncode element of the workflow.  I was unsure how to get this information.  Also, it isn't that straightforward since the evaluation takes place after the style and user prompt are merged and currently the job info only displays the user prompt.  If the style prompt has any use of template syntax, that wouldn't be visible anyway.  Not sure what to do with this, but it would be nice to know what actual prompt was used to generate an image that you want to refine or recreate.
- Similarly, even with a fixed seed, the generated image will be different on each generation due to random selection replacement of syntax.  This could actually be considered a feature.
- No json or yaml wildcard file support.  I could not find any documentation on how the other dynamic prompt plugins for comfy, a1111, or forge handle these files with their nested structure.  Majority of use cases I have seen are using txt files anyway.
- Cyclic or Combinatorial sampling would require state to persist between job queues, more UI elements, and even require automatic execution of batch queues.  This seemed out of scope for a simple feature and was not of particular interest to me.

### DIFF
--- a/ai_diffusion/text.py
+++ b/ai_diffusion/text.py
@@ -219,3 +219,43 @@ def edit_attention(text: str, positive: bool) -> str:
         if weight == 1.0 and open_bracket == "("
         else f"{open_bracket}{attention_string}:{weight:.1f}{close_bracket}"
     )
+
+
+def remove_comments_and_normalize(text: str | None) -> str:
+    """Remove Python-style comments and normalize whitespace."""
+    if text is None or text == "":
+        return ""
+
+    def trim_comments(line: str) -> str:
+        quote = None
+        for i, char in enumerate(line):
+            if char in "\"'":
+                quote = None if quote == char else char
+            elif char == "#" and not quote:
+                return line[:i]
+        return line
+
+    return "\n".join(
+        trimmed for line in text.split("\n") if (trimmed := trim_comments(line).strip())
+    )
+
+
+def find_matching_brace(text: str, start_pos: int) -> Tuple[int, bool]:
+    """Find matching closing brace, returning (end_pos, has_pipe)."""
+    depth = 1
+    has_pipe = False
+    pos = start_pos
+
+    while pos < len(text) and depth > 0:
+        if text[pos] == "{":
+            depth += 1
+        elif text[pos] == "}":
+            depth -= 1
+        elif text[pos] == "|" and depth == 1:
+            has_pipe = True
+        pos += 1
+
+    if depth > 0:
+        raise Exception(f"Unmatched opening brace at position {start_pos}")
+
+    return pos, has_pipe
--- a/ai_diffusion/dynamic_prompts.py
+++ b/ai_diffusion/dynamic_prompts.py
@@ -1,0 +1,344 @@
+import re
+import random
+from typing import Dict
+
+from .text import remove_comments_and_normalize, find_matching_brace
+from .util import client_logger as log
+from .wildcards import process_wildcards, find_wildcard_files, read_wildcard_file
+
+max_iterations = 100  # Maximum iterations for processing dynamic prompts
+max_recursion_depth = 10  # Maximum recursion depth for nested templates
+
+_var_name_pattern = r"[a-zA-Z0-9_]+"
+_var_ref_re = re.compile(rf"\${{({_var_name_pattern})(?::([^}}]+))?}}")
+_weight_re = re.compile(r"^(\d*\.?\d+)::(.*)")
+_wildcard_re = re.compile(r"^__([a-zA-Z0-9_*/]+)__$")
+_count_spec_re = re.compile(r"^(\d*-?\d*)(\$\$)")
+_immediate_var_re = re.compile(r"\${([a-zA-Z0-9_]+)=!")
+_regular_var_re = re.compile(r"\${([a-zA-Z0-9_]+)=")
+_inner_var_re = re.compile(r"^\${([a-zA-Z0-9_]+)=(.*)}$")
+_var_empty_re = re.compile(r"\${=")
+_separator_re = re.compile(r"\$\$(.*?)\$\$")
+
+
+def evaluate_dynamic_prompt(prompt: str) -> str:
+    """Process dynamic prompt templates and evaluate them."""
+    if not prompt:
+        return prompt
+
+    variables: Dict[str, str] = {}
+    prompt = remove_comments_and_normalize(prompt)
+    prev_prompt = ""
+    iteration = 0
+
+    while prev_prompt != prompt and iteration < max_iterations:
+        prev_prompt = prompt
+        iteration += 1
+        try:
+            prompt = _process_variables(prompt, variables)
+            prompt = process_variants(prompt)
+            prompt = process_wildcards(prompt, variables)
+        except Exception as e:
+            error_msg = f"Template processing error in iteration {iteration}: {str(e)}"
+            log.warning(error_msg)
+            raise Exception(error_msg) from e
+
+    if iteration >= max_iterations:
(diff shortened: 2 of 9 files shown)


Acly:
It's generally not possible to have dependencies for the plugin - it runs inside Krita's embedded python. I also feel like it's perhaps a bit overkill (or a lot) - I can see the usefulness of `{a|b|c}` and wildcard files, but not sure how many people actually need a template language with variable substitution etc.

Do you use all of it? Or can there be a "90% of the value with 10% of the code" kind of deal?

Acly:
Regarding where substitution happens, these are the stages:
1. `model.py` - collect all the data from the workspace
2. `workflow.py::prepare` - client-side processing of the data, results in a `WorkflowInput`
3. `workflow.py::create` - creates a comfyui prompt from `WorkflowInput`

Note that `WorkflowInput` must contain _all_ data which influences the workflow, and step 3. should not access globals (like settings) or the file system. That part will run on a different machine in cloud scenario.

Dynamic prompt evaluation could be applied in step 2 - this would also allow adding the result to job metadata.
* That's before merging style prompt, but I don't think having dynamic style prompt is useful/important?
* Needs to evaluate dynamic prompt for all regions

nolan778:
* Getting rid of variables would certainly cut down the code and headache, but unfortunately they are very useful, especially immediate evaluation variables that use the "!" symbol to force immediate evaluation so that the same randomly chosen word or phrase can be used in multiple places in your prompt.  Non immediate variables are also useful to shorten the visible length of the prompt when you don't want to list the same variant list or wildcard file multiple times.
* For the rest, I wanted the available syntax to be mostly the same to what people expect from the existing dynamic prompt plugins and they wouldn't have to change their saved prompts or modify their wildcard files to remove nested variables, variants, and references to other wildcard files.  I didn't see a major performance issue, but this plugin isn't trying to evaluate the prompt dozens of times per second and it is probably best to have this off for live mode anyway.
* if it helps, the dynamic prompt code could be moved to its own module to avoid cluttering the smaller text module.
* I was unaware about needing to do this in prepare() instead, but that makes sense and I will move the call(s) there.  I agree that we can probably go without a dynamic style prompt or a style prompt that influences the user prompt and vice versa.  Probably would not want that anyway, in case the user turned off dynamic prompts option and forgot to remove the syntax from their chosen style. Four questions through:
    - The positive prompt is modified a few times.  Where exactly in that function should I put it?
    - Why is the negative prompt merged immediately with the style negative, but the positive prompt is not?
    - I never learned what regions are or how to use them.  Should they share influence with the main prompt?  Would someone put a variable in the main prompt and use it in the regional prompt?  Or should the prompts be isolated and evaluated separately?  Sharing influence would really complicate it and require multiple calls to all prompts to iterate and pass a variable dictionary back and forth until all prompts stabilized.  Probably not what we want to do.
    - Should negative prompts even use dynamic prompts?  I'm thinking no.  I don't see the use case.  These days I pretty much only have a fixed list of negative tokens per model type and models don't seem to honor specific negative tokens related to concepts that they are dead set on depicting anyway.

Acly:
> The positive prompt is modified a few times. Where exactly in that function should I put it?

I think it's just handling of `<lora:...>` tags inside the prompt. They shouldn't clash with dynamic prompts, so order doesn't matter, but I'd do dynamic prompts afterwards.

> Why is the negative prompt merged immediately with the style negative, but the positive prompt is not?

In some cases positive prompt can affect only parts of the image while style prompt affects the whole image.

> I never learned what regions are or how to use them. Should they share influence with the main prompt? 

Regions are just prompts that affect certain part of the image. No I don't think state needs to be shared between region prompts.

> Should negative prompts even use dynamic prompts?

Personally I don't see the point, negative prompts have limited use. Doesn't stop some people from going ham on them though...

nolan778:
Okay, moving it to prepare() works as you said, but the job info is still showing the full prompt with template syntax instead of the final evaluated text.  What would I need to modify to capture it properly in the job history?

nolan778:
Okay, I think I finally got it working well in process_regions.  I added another string member to the Region and RootRegion classes to avoid modifying the original positive property directly and forcing a UI update with the evaluated text.  Then in process_regions, I evaluate the root and all regions and I use the evaluated positive property instead, making sure to clear the evaluation before returning from the function, so that the next generation starts from scratch.

ai_diffusion/text.py
```diff
@@ -219,3 +255,988 @@ def edit_attention(text: str, positive: bool) -> str:
...
+            raise Exception(error_msg)
+
+        # Select a random value
+        selected_value = random.choice(values)
+
+    # Apply parameters to the template
+    # First explicit parameters, then variables from context if needed
+    param_vars = params.copy()  # Create a copy to avoid modifying the original
+
+    # Apply any explicit parameters first (these take precedence)
```
Acly, line 1037: Just from skimming, lot's of duplication here...

(any experience if telling it to be "concise" or something avoids all those pointless comments?)

nolan778: Funny enough, asking models for a code review to find issues usually just results in extra comments instead of improved or fixed code.  I didn't have the tests while developing and was afraid to touch it once I got it all passing and working, due to the complex interactions between all the syntax.  With all the tests now, I can be more confident with a refactor.

nolan778:
Refactored text module to move all the dynamic prompt evaluation out to its own file and split it up even further to put wildcard handling in its own file as well.  This wildcard module may need to be expanded in the future to handle structured wildcard files (json and yaml), but that would require a ton of rework to how wildcard files are handled, so I'm not going down that road right now.

I also refactored the code as desired to be much more concise without breaking anything.

Next step is to investigate more on where to put the actual evaluation while still being able to capture it in the job history.  Without that info, there is no indication to the user what values were chosen.

PR #1735, opened 2025-04-27
Add IPEX support 

SHoogstad:
worked on this a while back just fineshed it let me now if anything needs to be changed!

### DIFF
--- a/ai_diffusion/server.py
+++ b/ai_diffusion/server.py
@@ -212,7 +212,11 @@ class Server:
         elif self.backend is ServerBackend.directml:
             torch_args = ["numpy<2", "torch-directml"]
         elif self.backend is ServerBackend.xpu:
-            torch_args += ["--index-url https://download.pytorch.org/whl/xpu", "intel-extension-for-pytorch==2.6.10+xpu","--extra-index-url https://pytorch-extension.intel.com/release-whl/stable/xpu/us/"]
+            torch_args += [
+                "--index-url https://download.pytorch.org/whl/xpu",
+                "intel-extension-for-pytorch==2.6.10+xpu",
+                "--extra-index-url https://pytorch-extension.intel.com/release-whl/stable/xpu/us/",
+            ]
         await _execute_process("PyTorch", self._pip_install(*torch_args), self.path, cb)
 
         requirements_txt = temp_comfy_dir / "requirements.txt"
--- a/ai_diffusion/server.py
+++ b/ai_diffusion/server.py
@@ -13,7 +13,13 @@ from . import eventloop, resources
 from .resources import CustomNode, ModelResource, ModelRequirements, Arch
 from .network import download, DownloadProgress
 from .localization import translate as _
-from .util import ZipFile, is_windows, create_process, decode_pipe_bytes, determine_system_encoding
+from .util import (
+    ZipFile,
+    is_windows,
+    create_process,
+    decode_pipe_bytes,
+    determine_system_encoding,
+)
 from .util import client_logger as log, server_logger as server_log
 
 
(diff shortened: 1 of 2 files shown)


Acly:
Hi, torch version was changed to 2.7 in latest update. It looks like the intel xpu extension needs to match the PyTorch version? Does it support 2.7, or does it make sense to allow a different version when installing for xpu?

side note: there seem to be some unrelated formatting changes, I'd prefer it if those could be removed from the PR

ai_diffusion/settings.py
```diff
@@ -24,6 +24,7 @@ class ServerBackend(Enum):
     cuda = (_("Use CUDA (NVIDIA GPU)"), not is_macos)
     mps = (_("Use MPS (Metal Performance Shader)"), is_macos)
     directml = (_("Use DirectML (GPU)"), is_windows)
+    xpu = (_("Use XPU(InteL gpu)"), is_windows)  # let me know if this is the correct way to do this
```
Acly, line 27: Looks fine, please change the string to `Use XPU (Intel GPU)` to be consistent with the rest

SHoogstad: jep will do!

SHoogstad:
> Hi, torch version was changed to 2.7 in latest update. It looks like the intel xpu extension needs to match the PyTorch version? Does it support 2.7, or does it make sense to allow a different version when installing for xpu?
> 
> side note: there seem to be some unrelated formatting changes, I'd prefer it if those could be removed from the PR

the torch and IPEX version do need to be match for best results i ran pre release of torch 2.7 with ipex 2.6 for a while but got strange bugs/errors with IPEX that why i choose a specific ver

SHoogstad:
also i am debating if whe should upgrade to 2.7 now its out for torch and ipex, work great i do see a speed decrease of around 2s (normally 22s now 24s)

Acly:
> edit: i would not reccomend it with a some testing i did ipex/torch 2.7 is a little unstable with performance atleast for intel

It's fine with me either way, I can't really judge since I don't have Intel GPU.

The other backends should stay on 2.7, so either the torch install needs to be conditionally changed to 2.6 when choosing XPU, or the intel-extensions moved to 2.7

SHoogstad:
> > edit: i would not reccomend it with a some testing i did ipex/torch 2.7 is a little unstable with performance atleast for intel
> 
> It's fine with me either way, I can't really judge since I don't have Intel GPU.
> 
> The other backends should stay on 2.7, so either the torch install needs to be conditionally changed to 2.6 when choosing XPU, or the intel-extensions moved to 2.7

forced it to go to 2.6 for intel/xpu

ai_diffusion/server.py
```diff
@@ -232,11 +232,30 @@ async def _install_comfy(self, comfy_dir: Path, network: QNetworkAccessManager,
             torch_args += ["--index-url", "https://download.pytorch.org/whl/cu128"]
         elif self.backend is ServerBackend.directml:
             torch_args = ["numpy<2", "torch-directml"]
+        elif self.backend is ServerBackend.xpu:
+            torch_args = [
+                "torch=2.6.0",
```
Acly, line 237: I think a double equals `==` is required for this to work?

SHoogstad: jep you right wrote it a little to fast

Acly:
Thanks for the PR!
Would be cool if you could bump the PyTorch version when you update in the future and find it to work well with a future release.

SHoogstad:
> Thanks for the PR! Would be cool if you could bump the PyTorch version when you update in the future and find it to work well with a future release.

2.6 is atm the most stable vers 2.7 just has some issues where you need to restart didn;t have that on 2.6

PR #1783, opened 2025-05-21
Update zh-cn.json

pokipoi:
Removed translations of embedded ads suspected to be Chinese online educational institutions. I hate that most educational institutions in China will claim that open source software is their own development, which is misleading to people


Acly:
Hm that's quite sneaky, thanks for letting me know

PR #1801, opened 2025-05-28
Add xpu to suported table and xpu minor performance increase for bigger canvas

SHoogstad:
Main purpose of this PR:
This PR introduces performance improvements to Krita AI specifically for Intel GPUs.

Performance comparison:

- Before this PR: ~66 seconds for 1920x1080 images

- After this PR: ~34–40 seconds for 1920x1080 images

### DIFF
--- a/ai_diffusion/workflow.py
+++ b/ai_diffusion/workflow.py
@@ -682,7 +682,11 @@ def scale_refine_and_decode(
         upscaler = models.upscale[UpscalerName.default]
 
     # if an canvas deviates both sizes from 1024 huge performance penalty tiled vae decreases it this is intel only
-    if extent.desired.width > 1536 or extent.desired.height > 1536 and Server.backend is ServerBackend.xpu:
+    if (
+        extent.desired.width > 1536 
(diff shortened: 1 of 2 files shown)


SHoogstad:
this also could benefit if its always on for the second pass if it isnt a 1024 x 1536 or 1024 x 1024

ai_diffusion/workflow.py
```diff
@@ -680,6 +681,14 @@ def scale_refine_and_decode(
         assert mode is ScaleMode.upscale_quality
         upscaler = models.upscale[UpscalerName.default]
 
+    # if an canvas deviates both sizes from 1024 huge performance penalty tiled vae decreases it this is intel only
+    if (
+        extent.desired.width > 1536
+        or extent.desired.height > 1536
+        and Server.backend is ServerBackend.xpu
```
Acly, line 688: `Server.backend` is a class field, as far as I know it remains constant (only the instance field is changed)

Using `settings.server_backend` should work. Generally should avoid accessing globals when building workflows, but since it's a local performance optimization it should work.

PR #1818, opened 2025-06-10
Improve XPU performance - Part 2 

SHoogstad:
This PR continues the work from [#1801](https://github.com/Acly/krita-ai-diffusion/pull/1801#event-18029226349).
After additional testing, I found that performance is significantly more stable when only the latest decode is tiled.


SHoogstad:
i do not know why cancel test is failing (it cant find krita from an import which is located in the docker folder)

Acly:
I don't really understand what this change is supposed to do - it adds another location to set tiled VAE, but doesn't remove the first one, so it should already be set anyway?

> i do not know why cancel test is failing (it cant find krita from an import which is located in the docker folder)

`root` can't be imported in workflows, there's a deliberate separation so workflows can test & run without depending on any UI or Krita stuff

SHoogstad:
> I don't really understand what this change is supposed to do - it adds another location to set tiled VAE, but doesn't remove the first one, so it should already be set anyway?

looks like sync added the old one back now its correct


> `root` can't be imported in workflows, there's a deliberate separation so workflows can test & run without depending on any UI or Krita stuff

how would i then get access to the device information from custom servers?

Acly:
> how would i then get access to the device information from custom servers?

By passing the required information into the functions that need it, like everything else. A clean way to do it might be to change the `tiled_vae` option which is passed as True/False at the moment to an enum with possible values `off`, `always`, `high_res`. Where high_res means when resolution is > 1536 and you pass that in when using xpu.

PR #1833, opened 2025-06-24
CLIP from model

moonwhaler:
Added a new setting (in advanced settings for a style) that allows to use the CLIP from the model. This is a boolean switch, which changes the way the workflow will be created (using a Load Checkpoint node) and connecting the model CLIP.


moonwhaler:
Not sure what happened to the diff here. My local diff only shows 22 lines of changes.

PR #1835, opened 2025-06-24
Get CLIP from model

moonwhaler:
Added a new setting (in advanced settings for a style) that allows using the CLIP from the model itself. This will use the "Load Checkpoint" node and connect the model CLIP. (Cleaned up my initial pull request.)

### DIFF
--- a/ai_diffusion/api.py
+++ b/ai_diffusion/api.py
@@ -64,6 +64,7 @@ class CheckpointInput:
     self_attention_guidance: bool = False
     dynamic_caching: bool = False
     tiled_vae: bool = False
+    use_model_clip: bool = False
 
 
 @dataclass
--- a/ai_diffusion/style.py
+++ b/ai_diffusion/style.py
@@ -84,6 +84,12 @@ class StyleSettings:
         _("Pay more attention to difficult parts of the image. Can improve fine details."),
     )
 
+    use_model_clip = Setting(
+        _("Use CLIP from Model"),
+        False,
+        _("Use the CLIP encoder which is bundled with the main model checkpoint."),
+    )
+
     preferred_resolution = Setting(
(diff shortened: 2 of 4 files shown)


Acly:
I don't quite understand what the purpose is here. As I've mentioned in the issue, you already get this behavior by putting your model file in the checkpoints folder where it belongs. What additional functionality would this switch unlock?

Also as far as I can see the switch would make _actual_ "diffusion models" unusable, since they do not have an embedded clip.

moonwhaler:
Okay, I will test this again and revoke this pull request, if this is true. In my local tests all models I used were fine with the switch being activated / deactivated. (Tested SDXL & Flux).

A part of my current "extra_model_paths.yaml" looks like this:

a111:
    base_path: /home/christian/AI/stable-diffusion-webui-forge/
    checkpoints: models/Stable-diffusion
    #diffusion_models: models/diffusion_models
    configs: models/Stable-diffusion
    vae_approx: models/VAE-approx
    vae: models/VAE
    unet: models/Stable-diffusion
    loras: |
         models/Lora
         models/LyCORIS
    upscale_models: |
                  models/ESRGAN
                  models/RealESRGAN
                  models/SwinIR
                  models/GFPGAN
                  models/EvTexture
    embeddings: embeddings
    hypernetworks: models/hypernetworks
    controlnet: models/ControlNet
    clip: models/text_encoder

Since I have one "central" folder for all checkpoints. In the mentioned checkpoints folder I have several subfolders, also a "PONY" subfolder. From there the model is fetched.

Acly:
I don't think  you should have `checkpoints` and `unet` pointing to the same folder. That might be that underlying issue here.

moonwhaler:
Okay. I feel very dumb right now. At least I learned a few things along the way. Thank you!
The change "fixed" it.

PR #1877, opened 2025-07-09
Magcache Support

dseditor:
Added support for MagCache. It has been tested locally link (custom server) with ComfyUI. MagCache is commonly used in video models such as FramePack and Wan, and it also supports models like Flux and Flux Kontext. This support includes updates to nodes and adjustments to detailed parameters.

https://github.com/Zehong-Ma/ComfyUI-MagCache

### DIFF
--- a/ai_diffusion/api.py
+++ b/ai_diffusion/api.py
@@ -40,7 +40,6 @@ class ImageInput:
     def from_extent(e: Extent):
         return ImageInput(ExtentInput(e, e, e, e))
 
-
 @dataclass
 class LoraInput:
     name: str
@@ -64,7 +63,12 @@ class CheckpointInput:
     self_attention_guidance: bool = False
     dynamic_caching: bool = False
     tiled_vae: bool = False
-
+    magcache_enabled: bool = False
+    magcache_thresh: float = 0.24  
+    magcache_retention_ratio: float = 0.1  
+    magcache_K: int = 5 
+    magcache_start_step: int = 0
+    magcache_end_step: int = -1
 
 @dataclass
 class SamplingInput:
--- a/ai_diffusion/ui/settings.py
+++ b/ai_diffusion/ui/settings.py
@@ -619,19 +619,57 @@ class PerformanceSettings(SettingsTab):
         self._dynamic_caching.value_changed.connect(self.write)
         self._layout.addWidget(self._dynamic_caching)
 
+        self._magcache_enabled = SwitchSetting(
+            Settings._magcache_enabled, 
+            text=(_("Enabled"), _("Disabled")),
+            parent=self
+        )
+        self._magcache_enabled.value_changed.connect(self.write)
+        self._magcache_enabled.value_changed.connect(self._update_magcache_advanced)
+        self._layout.addWidget(self._magcache_enabled)
+
+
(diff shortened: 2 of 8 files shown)


Acly:
Looks interesting.

I have two requests:
1. Show only one button (enable/disable) _OR_ one slider (strength). I don't think all the parameters should be exposed, unless there is a good reason to tweak them for certain scenarios. Default values for the models can be passed to the node.
2. There should be no need for `PerformanceSettings` in `workflow.py`? The parameter is already part of `CheckpointInput`. Please handle MagCache in the existing `load_checkpoint_with_lora` function (similar to WaveSpeed), don't introduce a new one.

ai_diffusion/workflow.py
```diff
@@ -153,6 +155,29 @@ def load_checkpoint_with_lora(w: ComfyWorkflow, checkpoint: CheckpointInput, mod
     if arch.supports_attention_guidance and checkpoint.self_attention_guidance:
         model = w.apply_self_attention_guidance(model)
 
+    # Apply MagCache as a model patch if enabled  
+    if checkpoint.magcache_enabled:
+        if arch in [Arch.flux, Arch.flux_k]:
+            print(f"Applying MagCache patch for {arch}")
```
Acly, line 161: Please remove all the prints

ai_diffusion/workflow.py
```diff
@@ -153,6 +155,29 @@ def load_checkpoint_with_lora(w: ComfyWorkflow, checkpoint: CheckpointInput, mod
...
+            
+            model_type = "flux_kontext" if arch is Arch.flux_k else "flux"
+            
+            try:
```
Acly, line 165: no need to wrap in try/except, this should not fail (or do so loudly if the code is broken)

ai_diffusion/ui/settings.py
```diff
@@ -640,12 +648,30 @@ def update_client_info(self):
...
                 + f": [{client.device_info.type.upper()}] {client.device_info.name} ({client.device_info.vram} GB)"
             )
-            self._dynamic_caching.enabled = client.features.wave_speed
-            self._dynamic_caching.setToolTip(
-                _("The {node_name} node is not installed.").format(node_name="Comfy-WaveSpeed")
-                if not client.features.wave_speed
-                else ""
-            )
+            
+            has_wave_speed = hasattr(client.features, 'wave_speed') and client.features.wave_speed
```
Acly, line 652: the `hasattr` is not needed

ai_diffusion/api.py
```diff
@@ -64,7 +63,12 @@ class CheckpointInput:
     self_attention_guidance: bool = False
     dynamic_caching: bool = False
     tiled_vae: bool = False
-
+    magcache_enabled: bool = False
+    magcache_thresh: float = 0.24
```
Acly, line 67: Please remove the parameters also from here and `settings.py`.

They can be dug up from git if ever needed, but I'd rather not have them in the code without a good reason.

And if more configurability is desired at some point, a single "strength" value or some preset would be a more user friendly way to control it anyway.

dseditor:
Thanks for Acly's detailed review. I'll try to address these issues. I think we might keep a threshold setting to accommodate models that will use MagCache in the future, but clean up unnecessary code

PR #1879, opened 2025-07-10
feat: add ctrl-click to cancel queued tasks before generating

FeepingCreature:
I notice myself a lot of times going "Queue->Cancel->Queued, Generate". I usually gen with 4 queued to keep my gpu busy, but often I can tell by the first generation that I did something wrong.

So, now Ctrl-click on the Generate button cancels queued tasks and requeues. (Thanks Sonnet 4!)

### DIFF
--- a/ai_diffusion/ui/generation.py
+++ b/ai_diffusion/ui/generation.py
@@ -761,6 +761,7 @@ class GenerationWidget(QWidget):
                 self.add_region_button.clicked.connect(model.regions.create_region_group),
                 self.region_prompt.activated.connect(model.generate),
                 self.generate_button.clicked.connect(model.generate),
+                self.generate_button.ctrl_clicked.connect(self.generate_replace),
             ]
             self.region_prompt.regions = model.regions
             self.custom_inpaint.model = model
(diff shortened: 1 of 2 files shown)


FeepingCreature:
Hm, it doesn't seem to always queue as many tasks as it should, or rather, it will queue four tasks and then only do one or two. Odd. Any idea what I'm doing wrong?

FeepingCreature:
Ah, cancel is async and the cancel and queue requests end up racing each other. I'm ... not sure how to fix that because there's multiple pieces in the code, ie. connection.py, model.py, where it just queues things and lets them run. Where *should* asyncness happen?

Acly:
> Where should asyncness happen?

It only _really_ happens in the client, where functions are suspended after making a HTTP request until sometime after there's a response.

But I think what you want to do is "first cancel existing jobs, then start new jobs" and the code doesn't do that.

ie. the difference between
```
model.cancel()  # no await
model.generate()
```
and
```
def generate_replace():
    await model._cancel()  # do one after the other
    await model._generate()

eventloop.run(generate_replace)
```
... where `_cancel` and `_generate` are the coroutine versions of the existing functions (ie they don't call `eventloop.run()` themselves)

as a rule, most code should use coroutines, and only the interface between UI / Logic (ie. `Model` and `Connection`) should wrap stuff in fire-and-forget tasks using eventloop. Just to be able to connect functions to Qt events.

Acly:
Maybe a better way to implement this would be similar to `Model.queue_front`

It could replace `queue_front` with some enum (`QueueMode.back/front/replace`) and be available in the same drop down? That would add a way to set "replace" for future normal-clicks. Can keep the Ctrl shortcut on top, although I might want to repurpose it at some point.

PR #1894, opened 2025-07-16
Provide prompt_id to ComfyUI, remove local/remote id split. (WAIT FOR COMFYUI)

FeepingCreature:
WARNING: This is merged in ComfyUI but not released yet. Seems to work in testing.

Feels like it should be possible to remove more stuff, but maybe that was before the recent refactor.

### DIFF
--- a/ai_diffusion/comfy_client.py
+++ b/ai_diffusion/comfy_client.py
@@ -35,24 +35,18 @@ if util.is_macos:
 
 @dataclass
 class JobInfo:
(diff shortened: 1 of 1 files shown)


Acly:
Looks good!

> Feels like it should be possible to remove more stuff, but maybe that was before the recent refactor.

I think the main issues were not directly related to the local/remote id split, it just provided some more opportunities for pitfalls.

PR #1895, opened 2025-07-16
feat: Switch to enum queue mode, add 'replace' mode.

FeepingCreature:
Follow up to https://github.com/Acly/krita-ai-diffusion/pull/1879

Kind of janky because of the fact that job queue tasks are split up before the client- I make clear-queue a property of the first job.

Not sure if the enum is in the right place.

Tentatively planning to implement ctrl-click as "Alternate queue mode".

### DIFF
--- a/ai_diffusion/model.py
+++ b/ai_diffusion/model.py
@@ -37,6 +37,12 @@ from .resources import ControlMode
 from .resolution import compute_bounds, compute_relative_bounds
 
 
+class QueueMode(Enum):
+    back = 0
+    front = 1
+    replace = 2
+
+
 class Workspace(Enum):
     generation = 0
     upscaling = 1
@@ -95,7 +101,7 @@ class Model(QObject, ObservableProperties):
     seed = Property(0, persist=True)
     fixed_seed = Property(False, persist=True)
     resolution_multiplier = Property(1.0, persist=True)
-    queue_front = Property(False, persist=True)
+    queue_mode = Property(QueueMode.back, persist=True)
     translation_enabled = Property(True, persist=True)
     progress_kind = Property(ProgressKind.generation)
     progress = Property(0.0)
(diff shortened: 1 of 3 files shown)


FeepingCreature:
The cleaner way would maybe be to allow queueing multiple jobs at once?

ai_diffusion/model.py
```diff
@@ -245,18 +246,30 @@ async def enqueue_jobs(
...
+        queue_mode = self.queue_mode
         for i in range(count):
             next_seed = sampling.seed + i * settings.batch_size
             input = replace(input, sampling=replace(sampling, seed=next_seed))
             params.seed = next_seed
-            job = self.jobs.add(kind, copy(params))
-            await self._enqueue_job(job, input)
+            new_job = self.jobs.add(kind, copy(params))
+            await self._enqueue_job(new_job, input, queue_mode)
+            if queue_mode is QueueMode.replace:
```
Acly, line 256: Managing the jobs here shouldn't be needed, the clients cancel the jobs and the job queue will react to reflect that.

Acly:
> The cleaner way would maybe be to allow queueing multiple jobs at once?

No I think this is fine.

Multiple jobs can be handled outside of clients in a client-agnostic way, so I wouldn't pull it in.

The replace logic can also be handled that way, ie. `enqueue_job` can call `client.clear_queue` if needed. Upside is that the entire replace logic is then in one place, and no need to modify both clients in the same way. On the other hand, queue-mode is passed to clients already anyway for "front", so the way you implemented also fits. I think both are fine.

Acly:
> Tentatively planning to implement ctrl-click as "Alternate queue mode".

a second drop-down sounds like overkill for just 3 options, I'd stick with fixed hotkeys

FeepingCreature:
Hm, if "stick with fixed hotkeys", should I just drop the dropdown option and inject it on ctrl-click?

edit: Or keep it in the dropdown and still inject on ctrl-click?

ai_diffusion/model.py
FeepingCreature: oh yeah huh, why did I not do that

Acly:
> Or keep it in the dropdown and still inject on ctrl-click?

yes keep it in the drop-down and also map it to ctrl+click

ai_diffusion/model.py
FeepingCreature: done

FeepingCreature:
kept in dropdown and added via ctrl-click

Acly:
Thanks!

PR #1907, opened 2025-07-27
Port to Krita 6 / PyQt 6

berolinux:
This makes krita-ai-diffusion work with current krita main branch, built with Qt 6.

### DIFF
--- a/ai_diffusion/client.py
+++ b/ai_diffusion/client.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 from abc import ABC, abstractmethod
 from enum import Enum
 from typing import Any, AsyncGenerator, Iterable, NamedTuple
-from PyQt5.QtCore import QObject, pyqtSignal
+from PyQt6.QtCore import QObject, pyqtSignal
 
 from .api import WorkflowInput
 from .image import ImageCollection
--- a/ai_diffusion/connection.py
+++ b/ai_diffusion/connection.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 from enum import Enum
-from PyQt5.QtCore import QObject, pyqtSignal, QUrl
-from PyQt5.QtGui import QDesktopServices
+from PyQt6.QtCore import QObject, pyqtSignal, QUrl
+from PyQt6.QtGui import QDesktopServices
 import asyncio
 
 from .client import Client, ClientMessage, ClientEvent, DeviceInfo, SharedWorkflow, MissingResources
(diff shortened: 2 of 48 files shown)


Acly:
This is interesting, fewer changes than I would've expected.

But I'm not sure what to do with the PR - I don't want to commit to maintaining two versions in parallel, and I think it's better to stick to official releases as that's what most people will be using.

AMDphreak:
This is an engineering failure in language architecture/design. You're right to avoid this headache...but I think we can make a build path that runs a script with rules in it. From what I can see, most of the edits here are simple string replacements, but there are a few functions that moved from one module to another.

@berolinux ~~Can you post a ruleset/specification here that satisfies this alternate build? I can think on how to make this run in the CI/CD script.~~

**Update:** I have now built a modular **Evolution Ecosystem** to solve this:
- **[Evolution Engine](https://github.com/dev-centr/equivalence-engine)**: A version-aware migration engine that parses SDL rules.
- **[Evolution Rules: Code](https://github.com/dev-centr/equivalence-rules-code)**: The repository for migration logic. The PyQt5 -> PyQt6 rules for this project are now live at `rules/python/PyQt`.

@Acly 

You can export a translated version of this project to a subfolder in order to build and release it separately. You can run the temporary migration for this project using:
`evolution-engine -p . --rules-repo https://github.com/dev-centr/equivalence-rules-code --library python/PyQt -f 5.15 -t 6.0`

I recommend sticking that line inside your CI build workflow to auto-build for you on release. To simplify CI builds on GitHub, you can use this action:
- **[Evolution Engine Action](https://github.com/dev-centr/equivalence-engine-action)**

## Permanent upgrade option

You can use `--in-place` or `-i` to update the original source code:
`evolution-engine --in-place -p . --rules-repo https://github.com/dev-centr/equivalence-rules-code --library python/PyQt -f 5.15 -t 6.0`

## Downgrading, or retro-fitting

`evolution-engine` can also **downgrade** code, so you can develop with Qt6 instead of Qt5 from now on. I see no reason to continue developing on Qt5.

TimofeyBiryukov:
Just ran in to an issue running the plugin with Krita 6 built from source which lead me here.
I had to downgrade to using Krita 5 via AppImage rather than building it which is not ideal for my use case.

From plugin manager in Krita 6 built from source:
image

```
from PyQt5.QtWidgets import QAction
ModuleNotFoundError: No module named 'PyQt5'
```

Would love to see this plugin working on latest version of Krita.

ThomasEricB:
> Just ran in to an issue running the plugin with Krita 6 built from source which lead me here. I had to downgrade to using Krita 5 via AppImage rather than building it which is not ideal for my use case.

Same issue.

r7vz9h3:
@ThomasEricB you can also try this:

https://github.com/Acly/krita-ai-diffusion/issues/2407#issuecomment-4214032492

Acly:
continued in #2491

Issue #660, opened 2024-04-25
(Suggesion) Saving images with the generation metadata


Miraihi:
Hello, Acly. By now I'm mostly satisfied with how you handle and advance the development of your plugin, thanks for your hard work!

Still, while it is not that critical of a problem, but it's still a bit of a hassle that the pictures generated don't contain any generation metadata Civitai can pick up. Is it even possible to implement?

Acly:
Hi! There are the pictures you generate 
* as they come from ComfyUI (potentially just partial masked images that have to be inserted and blended at the right offset)
* as they are saved by the history context menu
* whatever you save or export via Krita

The first would probably contain metadata if they were saved via ComfyUI, but end up somewhere in its output directory and be of limited use. The second I briefly tried to add metadata but it turned out Qt (which I use to write images) doesn't support writing the tags the way Comfy does, so it isn't much use.

As for what you export, that's usually the product of a lot of generations, so a bit unclear what it should contain?

Miraihi:
> As for what you export, that's usually the product of a lot of generations, so a bit unclear what it should contain?

At least the prompt and the negative prompt, but I've also seen the resources used (Loras) also saved as a part of the metadata. Usually they can be viewed using exiftool and built-in A111 image metadata reader.

Acly:
I meant in the case of saving/exporting Krita document, since it probably made of many prompts and prameters, it's unclear what to write.

The A1111 metadata is actually simple text, the only problem seems to be that it uses line breaks to separate the parameters. Somehow they aren't written properly, it seems like the image writer strips them.

Miraihi:
> it's unclear what to write.

Well, at least the metadata in one of the generated images you save by right clicking them in the plugin's window. A part of the prompt is already in the name of the file.

But as an alternative .txt sidecar with the metadata would also be acceptable.

Acly:
A1111 metadat looks something like this
```
prompt text goes here
Negative prompt: negative prompt text goes here
Steps: 20, Sampler: DPM++ 2M Karras, CFG scale: 7, Seed: 2341225840, Size: 512x512, Model: v1-5-pruned-emaonly
```
I can write that, but it's stripped of the line breaks, so comes out like this
```
prompt text goes hereNegative prompt: negative prompt text goes hereSteps: 20, Sampler: DPM++ 2M Karras, CFG scale: 7, Seed: 2341225840, Size: 512x512, Model: v1-5-pruned-emaonly
```
[...]

PR #1942, opened 2025-08-16
Add generation metadata to saved pngs from the generated thumbnails (from right-click or ... menu on thumbnails)

sfisher:
I hope this is helpful and I know at least a few other people are interested, but feel free to tell me to get lost if it's not something you can deal with and I'll just keep updating it for myself after every new release.

I find it useful to save the images from the generation thumbnails (select thumbnails and right-click or click "..." and *save image*).   Since I may come back to the PNGs later, it's very helpful to have some generation metadata in there (and currently none that I could see for files saved this way).

I've been running some code locally for about 4 months and haven't had any failures with thousands of images saved (also tested some UTF-8 characters like 4-bytle emoji and other characters in the metadata and it all seems to work).

I was originally going to use Pillow to do this, but there is a whole nightmare waiting with Pillow since it requires per-platform binaries and  makes things heavier, so found out that writing the headers into existing PNG files is relatively straight-forward so could just include a method to do it without all the overhead.

I added most of the code to the util class, but don't know if it belongs there or elsewhere (though they are metadata/file rewriting utils).

Trying to catch a plethora of exceptions so if something goes wrong it just saves the file and doesn't save the metadata and also tried to make minimal changes to existing code.

I looked into contributing guidelines, added some tests, ruff passes.

I think I had some trouble running tests which may require additional server setup or something, sorry I'm a bit lost about how to do that.

So I don't know if this interferes with any of the other tests but the util tests pass, since I'm not sure I have a complete testing setup.

LMK what you think and I'm happy to try to make additional changes if this is something you're interested in including.  I think it makes the PNGs a lot nicer so you have some idea how they were generated.


Acly:
We can include it. There's a few changes I'd like to see:
* Add an option in Interface settings to enable/disable metadata. There are some people who don't want their "prompts leaked" when they share images... see #660 
* Instead of writing PNG to file, then reading the file back into memory, use `Image.to_bytes` to do it directly
* Reorganize the code a bit:
  * `image.py`: put image saving code here, maybe `Image.save_with_metadata(metadata: str)` or similar
  * `text.py`: the code that converts job info into metadata could be put here, it already has lots of similar code

ai_diffusion/model.py
```diff
@@ -1391,4 +1393,23 @@ def _save_job_result(model: Model, job: Job | None, index: int):
     base_image = model._get_current_image(Bounds(0, 0, *model.document.extent))
     result_image = job.results[index]
     base_image.draw_image(result_image, job.params.bounds.offset)
+
+    # save the image to the file and then add the generation metadata afterwards by resaving to avoid having to load Pillow or other image libraries
     base_image.save(path)
+    try:
+        # now resave the image to add the generation metadata
+        util.resave_image_with_metadata(img_path=path, job=job)
+    except (
```
Acly, line 1402: `except Exception as e:` is good enough

ai_diffusion/util.py
```diff
@@ -296,3 +299,110 @@ def acquire_elements(l: list[QOBJECT]) -> list[QOBJECT]:
         if obj is not None:
             sip.transferback(obj)
     return l
+
+
+# Resave the image with PNG metadata
+def resave_image_with_metadata(img_path: Path, job: "Job"):  # noqa: F821
```
Acly, line 305: I don't think this function or the explanation in comments is needed.

Also if the function only needs the `params` from the job, pass that in directly instead of the entire job.

sfisher:
Thanks for the good feedback.  I'll see what I can do.  It may take me a little time because of some other things going on, but it's very helpful.

ai_diffusion/model.py
sfisher: With refactor, this is now gone

ai_diffusion/util.py
sfisher: Change to just pass in the JobParams

sfisher:
Ok.  I did a round of refactoring.  Have the PNG being saved from memory rather than writing to the file system and added the setting to save the metadata under the  "interface" options, added translations and moved tests around and made sure they pass as well as linting with ruff, etc.

It all seems to work fine to me.  Give it a try and see what you think.

I decided not to make full PNG writing code from scratch since there is already code to create PNGs that handles any extra complications about different color formats, compression, etc, and figured it was easier to leave it alone to just add the headers needed for the metadata.

Acly (commented):
Changes look good, I have a few smaller suggestions, then this should be ready

Sil3ntKn1ght:
i certainly don't want my meta in my images, I'll spend a week on some prompt tuning etc and the output image is a passion of time and love, iv had works from other output taken and passed of as there own. that's the worst feeling in the world.

if implemented the default should be off, for myself i save the .krz for work flows and data information.
i don't want it in png, jpg, etc

sfisher:
This is not something you need to worry about.  The default is off but can be changed to suit a user's preferences.  There are both reasons to want and reasons not to want to export this, which I understand.

Personally, I may create other variations and different prompts that may not be saved within the KRZ file and it's useful for me to be able to track the generations I export. Other people enjoy sharing some of their process or have other reasons to want or not want this.

I also have additional workflows (scaling/image enhancement) where this information is not passed along.

Be aware this ONLY applies to saving a generated png directly from the generation thumbnails and not exporting through the Krita File > Export menu which is maybe what you're doing if you're spending weeks.  It also follows the preference (off is default) as mentioned above which does not change the current behavior unless changed by the user.

ai_diffusion/image.py
```diff
@@ -8,8 +8,12 @@
 from pathlib import Path
 
 from .settings import settings, ImageFileFormat
+print("ImageFileFormat defined:", "ImageFileFormat" in globals())
```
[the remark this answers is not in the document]
sfisher: Oops.  I guess I missed that.

ai_diffusion/language/es-ca.json
```diff
@@ -311,6 +311,7 @@
     "Sampler": "Mostrejador",
     "Sampler Settings": "Ajustos del mostrejador",
     "Sampler Steps": "Passos del mostrejador",
+    "Save Automatic1111-Style Metadata": "Guardar metadades d'estil Automatic1111",
```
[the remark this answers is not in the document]
sfisher: Reverted all these files.

sfisher:
I believe this is ready since I made those changes, fixed a test and everything is still working afaict.  Thanks.  This will be nice to have the option without having to maintain my own branch or other things.

Streamtabulous:
i certainly don't want my meta in my images, I'll spend a week on some prompt tuning etc and the output image is a passion of time and love, iv had works from other output taken and passed of as there own. that's the worst feeling in the world.

if implemented the default should be off, for myself i save the .krz for work flows and data information.
i don't want it in png, jpg, etc

Streamtabulous:
> This is not something you need to worry about. The default is off but can be changed to suit a user's preferences. There are both reasons to want and reasons not to want to export this, which I understand.
> 
> Personally, I may create other variations and different prompts that may not be saved within the KRZ file and it's useful for me to be able to track the generations I export. Other people enjoy sharing some of their process or have other reasons to want or not want this.
> 
> I also have additional workflows (scaling/image enhancement) where this information is not passed along.
> 
> Be aware this ONLY applies to saving a generated png directly from the generation thumbnails and not exporting through the Krita File > Export menu which is maybe what you're doing if you're spending weeks. It also follows the preference (off is default) as mentioned above which does not change the current behavior unless changed by the user.

thanks, where does right clicking the generated thumb and saving it go to, i saved but can't find it to test it, iv been away for a year lot of new stuff

sfisher:
When saving generated images from the thumbnails they are saved in the same directory where the original file is saved.  I believe the option is grayed out if the file is not saved anywhere (such as a new image not saved yet).

Acly:
Thanks, works fine for me

One thing I noticed is that it starts metadata with "Prompt: ", whereas A1111 starts with the actual prompt directly (at least some 1+year old images I had lying around do). It doesn't particularly matter to me, but maybe it won't be read correctly in some viewers.

sfisher:
Thank you for catching that. I could do a tiny PR to change that or you're welcome to change it if you'd rather.

Thanks for your help!

Acly:
changed it: https://github.com/Acly/krita-ai-diffusion/commit/11cd88817d26d733ddaa3aadc909735eb17e7068

nolan778:
Was it expected that this would not include the style influence on the metadata saved to the PNG?  Seems like all that is missing, meaning the workflow opened in ComfyUI using the image cannot reproduce Krita AI Diffusion images.

Issue #1813, opened 2025-06-08
Error during server startup: RuntimeError: Failed to load the backend extension: intel_extension_for_pytorch. You can disable extension auto-loading with TORCH_DEVICE_BACKEND_AUTOLOAD=0. [1]


jojoklang:
Here are some logs i think that are related to it

```
2025-06-09 07:38:34,750 WARNING Failed to launch/connect server at startup: Error during server startup: RuntimeError: Failed to load the backend extension: intel_extension_for_pytorch. You can disable extension auto-loading with TORCH_DEVICE_BACKEND_AUTOLOAD=0. [1]
2025-06-09 07:38:41,516 INFO Using uv: uv 0.6.10
2025-06-09 07:38:41,696 INFO Using Python: Python 3.12.3, /home/user/.local/share/krita/ai_diffusion/server/venv/bin/python
2025-06-09 07:38:41,696 INFO Starting server with python -su main.py
```

jojoklang:
I use ubuntu latest 
Intel cpu
I installed pytorch and the other torch dependencies but using a venv as i don't know how to install system wide

Acly:
If you want to use CPU, select CPU.
The XPU option is for supported Intel Arc GPUs mainly, if you don't have one it will fail to load.

jojoklang:
Oh.. I missed that, when installing XPU was selected, even with CPU the error is the same, I thought that maybe unnecesary components were installing as XPU was selected and that could mess with the lauch. 

I am trying a re-install now with CPU selected. But if there is anything I don't know please tell me.

jojoklang:
It didn't work. I had to delete the the entire plug-in even the folder then start again. The server is now working. Thanks.

Acly:
I'll try to add `TORCH_DEVICE_BACKEND_AUTOLOAD=0` as the error message suggests when the server is launched in CPU mode but has been installed with XPU. Maybe it'll allow it to work in the future.

PR #1957, opened 2025-08-22
bump torch to 2.8 for xpu

SHoogstad:
I bumped the torch ver to 2.8 as is quite stable (in my testing) and remove ipex as its not recommended anymore as its causes the gen times to be longer + all optimaztions it has is already in pytorch


Acly:
Looks good. This means the workaround for https://github.com/Acly/krita-ai-diffusion/issues/1813 is no longer needed I guess, maybe it should be removed

SHoogstad:
jup 
will add it to the pr

Acly:
Thanks!

PR #1992, opened 2025-09-01
Add Options: Save Image Format choice (PNG/WebP/JPEG), Image Quality, Image Name Format

nolan778:
I often have to save many images and then run them through batch converters and renaming scripts, so I added some options to remove this step for myself.

New Options:

- Save Image Format (Default PNG (fast)) - Can now choose PNG (fast), PNG, WebP, WebP (Lossless) and JPEG.  This changes which format is used for right click context Save Image in job history thumbnails.
- Save Image Quality WebP (Default 80) - Hidden option only in settings json file.  This will override the default image quality settings in the save function for lossy webp image formats.
- Save Image Quality JPEG (Default 85) - Hidden option only in settings json file.  This will override the default image quality settings in the save function for jpeg image formats.
- Save Image File Name Template - Hidden option only in settings json file.  The default value is the current naming format, but the user can now modify this.  I added {current_timestamp} as another possible key, but there may be more in the future.

Modified:

- Save Image Metadata option is now hidden when neither PNG image save format is chosen.

I had to pass around optional format and quality to the save function and optional quality to the write function.  The save function used the file extension to determine format, but we can now be saving 2 different png and 2 different webp formats.  The write function used the default quality for each format, but the user may modify these qualities for lossy webp and jpeg by editing the settings.json file.

I couldn't easily figure out how to save the ComfyUI workflow metadata to the jpeg and webp formats, so i left this to someone more qualified than myself.  That option remains PNG only, at this time.

### DIFF
--- a/ai_diffusion/image.py
+++ b/ai_diffusion/image.py
@@ -522,11 +522,14 @@ class Image:
         array = np.frombuffer(ptr, np.uint8).reshape(h, w, c)  # type: ignore
         return array.astype(np.float32) / 255
 
-    def write(self, buffer: QIODevice, format=ImageFileFormat.png):
+    def write(
+        self, buffer: QIODevice, format=ImageFileFormat.png, override_quality: int | None = None
+    ):
         # Compression takes time for large images and blocks the UI, might be worth to thread.
         if not qt_supports_webp():
             format = format.no_webp_fallback
-        format_str, quality = format.value
+        format_str = format.extension
+        quality = override_quality if override_quality is not None else format.quality
         writer = QImageWriter(buffer, QByteArray(format_str.encode("utf-8")))
         writer.setQuality(quality)
         result = writer.write(self._qimage)
@@ -573,18 +576,25 @@ class Image:
         painter.drawImage(*offset, image._qimage)
         painter.end()
 
-    def save(self, filepath: Union[str, Path]):
-        fmt = ImageFileFormat.from_extension(filepath)
+    def save(
+        self,
+        filepath: Union[str, Path],
+        format: ImageFileFormat | None = None,
+        quality: int | None = None,
+    ):
+        fmt = format or ImageFileFormat.from_extension(filepath)
         file = QFile(str(filepath))
         if not file.open(QFile.OpenModeFlag.WriteOnly):
             raise Exception(f"Failed to open {filepath} for writing: {file.errorString()}")
         try:
-            self.write(file, fmt)
+            self.write(file, fmt, quality)
         finally:
             file.close()
 
-    def save_png_with_metadata(self, filepath: Union[str, Path], metadata_text: str):
-        png_bytes = bytes(self.to_bytes(ImageFileFormat.png))
+    def save_png_with_metadata(
+        self, filepath: Union[str, Path], metadata_text: str, format: ImageFileFormat | None = None
+    ):
+        png_bytes = bytes(self.to_bytes(format or ImageFileFormat.png))
         self.save_png_w_itxt(filepath, png_bytes, "parameters", metadata_text)
 
     def debug_save(self, name):
--- a/ai_diffusion/settings.py
+++ b/ai_diffusion/settings.py
@@ -69,11 +69,11 @@ class PerformancePreset(Enum):
 
 
 class ImageFileFormat(Enum):
-    png = ("png", 85)  # fast, large files
-    png_small = ("png", 50)  # slow, smaller files
-    webp = ("webp", 80)
-    webp_lossless = ("webp", 100)
-    jpeg = ("jpeg", 85)
+    png = "PNG (fast)"  # fast, large files
+    png_small = "PNG"  # slow, smaller files
+    webp = "WebP"
+    webp_lossless = "WebP (lossless)"
+    jpeg = "JPEG"
 
     @staticmethod
     def from_extension(filepath: str | Path):
@@ -82,9 +82,33 @@ class ImageFileFormat(Enum):
             return ImageFileFormat.png_small
         if extension == ".webp":
             return ImageFileFormat.webp
-        if extension == ".jpg":
+        if extension == ".jpg" or extension == ".jpeg":
             return ImageFileFormat.jpeg
         raise Exception(f"Unsupported image extension: {extension}")
+
+    @property
+    def extension(self):
+        if self in [ImageFileFormat.png, ImageFileFormat.png_small]:
+            return "png"
+        elif self in [ImageFileFormat.webp, ImageFileFormat.webp_lossless]:
+            return "webp"
+        else:
+            return "jpg"
+
+    @property
+    def quality(self):
+        if self in [ImageFileFormat.png]:
+            return 85
+        elif self in [ImageFileFormat.png_small]:
+            return 50
+        elif self in [ImageFileFormat.webp]:
+            return 80
+        elif self in [ImageFileFormat.webp_lossless]:
+            return 100
+        elif self in [ImageFileFormat.jpeg]:
+            return 85
+        else:
+            return 85
 
     @property
     def no_webp_fallback(self):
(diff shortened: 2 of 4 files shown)


ai_diffusion/settings.py
```diff
@@ -58,6 +58,12 @@ class ApplyRegionBehavior(Enum):
     no_hide = _("Layer group (don't hide)")
 
 
+class SaveImageFormat(Enum):
```
Acly, line 61: Rather than introducing a new Enum I think it would be better to use the existing `ImageFileFormat`.

You can adapt it to fit the settings UI if needed, eg. could look like this
```python
class ImageFileFormat(Enum):
    png = "PNG (fast)"
    png_small = "PNG"
    webp = "WebP"
    webp_lossless = "WebP (lossless)"
    jpeg = "JPEG"

    @property
    def extension(self):
        if self in [ImageFileFormat.png, ImageFileFormat.png_small]:
            return "png"
        elif self in [ImageFileFormat.webp, ImageFileFormat.webp_lossless]:
            return "webp"
        else:
            return "jpg"

    @property
    def quality(self):
        ...
```

ai_diffusion/settings.py
```diff
@@ -217,9 +223,40 @@ class Settings(QObject):
 
     save_image_metadata: bool
     _save_image_metadata = Setting(
-        _("Save Image Metadata"),
+        _("Save Image Metadata (PNG)"),
```
Acly, line 226: Maybe in the settings UI, disable this setting when something other than PNG is selected.

I'd keep the name and description unchanged in that case, makes it less likely to require retranslation.

ai_diffusion/settings.py
```diff
@@ -217,9 +223,40 @@ class Settings(QObject):
...
+    )
+
+    save_image_format: SaveImageFormat
+    _save_image_format = Setting(
+        _("Save Image Format"),
+        SaveImageFormat.png,
+        _("File format for saved images from thumbnails."),
+    )
+
+    save_webp_lossless: bool
```
Acly, line 238: I think both lossless and quality setting should be rolled into the format. Ie. just have "Webp" and "WebP (Lossless)" options instead of an extra toggle. Similar to quality slider if needed at all. I think we can make some assumptions here and pick good defaults. It's actually harder with one slider for all formats to pick a good default value.

ai_diffusion/ui/settings.py
```diff
@@ -487,6 +488,32 @@ def __init__(self):
...
+        self._widgets["save_webp_lossless"].value_changed.connect(
+            self._update_image_quality_widgets
+        )
+
+        add_header(self._layout, Settings._save_image_file_name_format)
+        self.add(
+            "save_image_file_name_format",
+            TextSetting(Settings._save_image_file_name_format, parent=self),
+        )
+        self._widgets["save_image_file_name_format"].show_label = False
```
Acly, line 505: I'd prefer to keep this setting editable in the `settings.json` for now (ie. you can keep using it, but without UI). It feels quite niche to me and not so easy to use, and I'm trying to avoid interface settings from growing too much. Probably also not something you constantly need to change.

ai_diffusion/ui/settings.py
```diff
@@ -496,6 +523,10 @@ def __init__(self):
 
         self._layout.addStretch()
 
+    def read(self):
+        super().read()
+        self._update_image_quality_widgets()
+
```
Acly, line 529:
```suggestion
    def _read(self):
        self._update_image_quality_widgets()

```

ai_diffusion/image.py
```diff
@@ -527,6 +527,9 @@ def write(self, buffer: QIODevice, format=ImageFileFormat.png):
             format = format.no_webp_fallback
         format_str, quality = format.value
         writer = QImageWriter(buffer, QByteArray(format_str.encode("utf-8")))
+        if format is ImageFileFormat.webp or format is ImageFileFormat.jpeg:
```
Acly, line 530: `image` should not change behavior depending on global UI settings, since it's used in a lot of places, while the setting is only supposed to affect the history image save

but I think this can be resolved by pulling all the necessary information into the `ImageFileFormat`, see other comments

ai_diffusion/settings.py
nolan778: Actually, I notice that the images are stored as webp inside the history itself.  Does that mean a higher quality when saving the images is irrelevant?  If 80% default quality is used to store the images before they are saved, you would just be making larger files with no increased quality benefit.  That also means increasing the webp default image quality would balloon the document size/memory usage, right?  That was unclear to me as a user and probably to others who probably used the normal Krita File Export options to save an image with high quality settings and didn't see any benefit.

nolan778: Yea, it seems that way.  I'm just gonna scrap the whole quality thing and run a script to modify the default quality in settings.py after each plugin update.  I do want higher quality, but that would be detrimental to the memory usage for others and you already don't put the image history format in the UI, so having a quality setting for images that are already low quality in the history seems silly.

Acly: Images are stored lossless in memory as long as you have the document open.
Compressed versions are stored to the .kra file. So you only loose quality if you save & open again. Choosing a higher quality still makes sense for the Save Image.

(and of course any image you apply from history is transferred losslessly to Krita layers, and can be exported with Krita's export in all sorts of formats with many quality sliders)

ai_diffusion/ui/settings.py
nolan778: This line is necessary or the UI breaks.  Took me a bit to figure out what was causing that after taking this change.

ai_diffusion/settings.py
nolan778: Okay, thanks for the clarification.  In that case, I have brought back the quality setting, but made it two settings (one for each jpeg/lossy webp) and I hide them appropriately.  Lossy webp option removed and integrated to the existing format.

ai_diffusion/image.py
nolan778: Removed the direct settings influence in this function, but I had to add an optional override quality to the function to allow the save function to change the quality away from the default on each format.  This was necessary for lossy webp and jpeg only.

ai_diffusion/ui/settings.py
Acly: you tried with `_read` (mind the underscore, it's a different function)?

ai_diffusion/image.py
```diff
@@ -572,13 +575,33 @@ def draw_image(self, image: "Image", offset: tuple[int, int] = (0, 0), keep_alph
         painter.drawImage(*offset, image._qimage)
         painter.end()
 
-    def save(self, filepath: Union[str, Path]):
-        fmt = ImageFileFormat.from_extension(filepath)
+    def save(self, filepath: Union[str, Path], format: ImageFileFormat | None = None):
+        fmt_from_ext = ImageFileFormat.from_extension(filepath)
+        fmt = fmt_from_ext
+        if format is not None and fmt_from_ext is not format:
```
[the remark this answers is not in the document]
Acly: I think the explicit format should take precedence, file extension is just a convention. But sounds like a bug anyway if that happens. Could add
```
assert format.extension == ImageFileFormat.from_extension(filepath).extension
```
but not sure if it breaks something, doesn't seem worth the resik

nolan778:
There is a conflict since you moved a nearby line.  After rebasing and resolving the conflict, the plugin no longer launches for me.  I realized there was also a new file and I got that added, but it still will not launch.  There are never any error messages in the client.log file that tell me why the plugin doesn't launch, when something goes wrong.  Krita.log also doesn't indicate any issues.  Where can I look to find the issue?

EDIT: Fixed the issue (was missing a symbolic link to another changed file), but it would be good to know how to debug issues going forward.  So far I have had to resort to comment out half the remaining code changes until it works again and do that un-commenting half of that, doing a manual binary search to find problematic lines of code.

Acly:
looks good now, will merge later or tomorrow

> Fixed the issue (was missing a symbolic link to another changed file), but it would be good to know how to debug issues going forward. 

not sure why you're sumlinking individual files, I think you just want to symlink the entire ai_diffusion folder

this covers most of it: https://github.com/Acly/krita-ai-diffusion/blob/main/CONTRIBUTING.md

One thing that isn't mentioned, on Windows, to get terminal output from Krita, you need to run `krita-install-dir/bin/krita.com` (NOT krita.exe). If the plugin doesn't load it will print error & trace to the terminal. You can also use prints in code, or attach debugger.

And running `pyright` will catch syntax problems, don't have to run Krita for that.

PR #2002, opened 2025-09-04
Prevent job history misclick on new image row

nolan778:
### Conditions:
* Image generation jobs already in progress and the results are coming in every few seconds, adding images to the job history UI
* The incoming image added causes the job history to scroll down (new job category or images wrapping to next line)
* Right clicking on an image within a fraction of a second after the latest image came in and caused a scroll
* Right click selects the new image in the new row when the user intended to select an image in the previous row.
### More Issue Details:
* Even though the image in the new row was selected by the right click, the highlight in the UI is sometimes still incorrectly on the old selected image.  This is not the case if the user waits a bit more time before right clicking the image in the history.  It only occurs in a small window of time where new images added.
* When selecting to delete the image from the opened context menu in this scenario, the new image gets deleted, even though the highlight is on the old image and the user thinks they are deleting the old image.
* The problem gets progressively worse the longer the job history gets, as I suppose there is some additional lag involved with adding new images at that point

### Fix discussion:
* Even if you could fix the highlight going on the wrong image, the user may be clicking through the right click context menu quickly enough to not stop their intended delete action before noticing the wrong image was going to deleted. I have done this dozens of times.
* You could disable all click actions and context menu open action on the job history UI for a a small window of time (300 ms) after a new image is added to the job history UI.  However, that happens when any images is added and not just when you are likely to mis-click (auto scroll).
* Better to detect when window scroll is likely to cause mis-click.
* Therefore, I think the best workaround solution is to disable all click actions and context menu open action on the job history for a small window of time (300 ms) after the job history window auto scrolls to the bottom and the scroll range changed.  If the user clicks any image during that window of time, no action occurs.  No image is selected and no context menu is opened.  The user must try to click again after the time window has passed.

This does fix the issue for me, but the solution is a little ugly, requiring 3 different timers.  Two to detect when the scroll to bottom and scroll range changed at the same time and one to guard against mis-clicking after that condition is detected.  If you have a more elegant or simpler solution to this problem, let me know.

### DIFF
--- a/ai_diffusion/ui/generation.py
+++ b/ai_diffusion/ui/generation.py
@@ -33,6 +33,8 @@ class HistoryWidget(QListWidget):
     item_activated = pyqtSignal(QListWidgetItem)
 
     _thumb_size = 96
+    _scroll_recent_ms = 10
+    _misclick_guard_ms = 300
     _applied_icon = Image.load(theme.icon_path / "star.png")
     _list_css = f"""
         QListWidget {{ background-color: transparent; }}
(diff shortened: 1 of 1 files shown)


Acly:
The highlight not being at the right position could be improved. But I think to fix it properly means trying to rewrite the buttons displayed on top of thumbnails using perhaps QStyledItemDelegate subclass. This is a lot of work tweaking everything to look right on all sizes / thumbails aspect ratios / screen DPI etc. I really don't think it's worth it.

Regarding misclicks: if you keep your history not scrolled to the bottom it won't auto-scroll, so no chance of misclicks. Or just wait until generation is done before deleting stuff.

I admit I don't understand the need to constantly micro-manage the history in the first place, it's more like browser history. The discard option was mostly because people wanted to occasionally delete accidental / offensive generations.

IMO this PR adds too much extra complexity and things that can go wrong, and I can imagine it being more annoying than helpful.

PR #2052, opened 2025-09-23
Update download_models.py

thc282:
the function in resources.py is changed
from file_id => verify

### DIFF
--- a/scripts/download_models.py
+++ b/scripts/download_models.py
@@ -104,10 +104,10 @@ def list_models(
     # Remove duplicate files listed under different IDs (apply to multiple architectures)
     file_ids = set()
     for m in list(models):
-        if m.file_id in file_ids:
+        if m.verify in file_ids:
             models.remove(m)
         else:
-            file_ids.add(m.file_id)
+            file_ids.add(m.verify)
 
     if len(models) == 0:
         print("\nNo models selected for download.")


Acly:
What was the issue exactly?

The change you made won't work for filter duplicates (and it doesn't even really make sense), so I'm wondering what was the idea behind it

thc282:
im trying with using local & host comfyui
the local works fine, just choose and download
but when i use custom hosting, i try to use this .py & Following the handbook webpage 
(just to download the models inside comfyui)

When i tried to run the py inside
`<appdata>/krita/pykrita`
it give me an error, since this py is using the package inside `ai_diffusion` folder

at last, i found that one of the function name already changed. so it will throw the error and the program will stop.

after correcting the right function name from `file_id()` to `verify()`. the py file run with no no issue

Acly:
I think there was an issue with using `download_models.py` from the `ai_diffusion` folder. I tried to fix it here: https://github.com/Acly/krita-ai-diffusion/commit/6a059f97c69e67157f886d5f2362ea995939f65a

I'm not sure it was the same issue you saw though. Because the `file_id` was not a problem.

Using `verify` instead of `file_id` is not a fix, they are completely different things, and the filtering will not work.

thc282:
great, now the issue fix
it is the name problem issue

PR #2076, opened 2025-10-09
Qwen Styles/Icons/Cleanup

Acly:


Danamir:
Nice, I'll do some tests this evening to see if everything is alright.

Danamir:
It seems all right !

Issue #2082, opened 2025-10-11
Text encoders / VAE models  (Qwen, Flux) not found in 1.40.0


JasonS09:
Hello, I have installed Qwen according to: https://github.com/Acly/krita-ai-diffusion/discussions/2081

However the model is not found.

Here is my diffusion_models folder:

<img width="1038" height="380" alt="Image" src="https://github.com/user-attachments/assets/c247e550-2f34-4038-90cb-a9935810b527" />

Here is my text_encoders folder:

<img width="784" height="261" alt="Image" src="https://github.com/user-attachments/assets/716d1d4e-58d9-4c20-bcdb-4fbabef57aaa" />

This shows when I connect to the server:

<img width="951" height="424" alt="Image" src="https://github.com/user-attachments/assets/f414a93d-af62-4e58-b6d8-cfb9ed1966f3" />

nakanakapoi:
Also happen to me, already have the models and updated everything. Hoping for fix soon.

Acly:
Fixed here: https://github.com/Acly/krita-ai-diffusion/commit/785c26f86abb7dd0183b59b3f47f61867458eb76

Possible workarounds (either one should work):
* install [ComfyUI-GGUF](https://github.com/city96/ComfyUI-GGUF) nodes (just need the nodes, no extra models)
* use latest version of the plugin from GIT
* apply the patch above manually
* revert to previous plugin version

shadowlocked:
> install [ComfyUI-GGUF](https://github.com/city96/ComfyUI-GGUF) nodes (just need the nodes, no extra models)

This was what worked for me.

pmads87:
After updating to 1.4.0 I get this when trying to load Flux:
`size mismatch for shared.weight: copying a param with shape torch.Size([256384, 4096]) from checkpoint, the shape in current model is torch.Size([32128, 4096]).`

Also, for SDXL and Illustrious, clip_g seem to be missing all of a sudden:
`Exception: Text Encoder 'clip_g' for SD XL not found`
`Exception: Text Encoder 'clip_g' for Illustrious not found`

I tried manually downloading [this version ](https://huggingface.co/Comfy-Org/stable-diffusion-3.5-fp8/blob/main/text_encoders/clip_g.safetensors) of clip_g, placing it in /text_encoders/clip_g.safetensor. Now the model loads and generates, but the images look... horrendous.

For me, only Qwen works.

momokyara:
[...]

PR #2084, opened 2025-10-11
Avoid replacing found text encoder models with None in GGUF model search

v0xie:
Updating the available_resources dict will replace the values with None if no GGUF model is found. This PR makes it so the values will only be updated if a model is found.

Feel free to close this PR if there's a better solution.  Thanks!

Related issues: #2082, #2072


Acly:
Thanks, I probably missed this because all my tests had the GGUF nodes installed.

PR #2108, opened 2025-10-17
Add Turkish README file

alperktt:
Maybe we can add turkish read me or other languages tooo


Acly:
This PR seems to delete the original README?

But I'm sceptical about adding translated README in general. Unlike the plugin translation where incomplete parts will simply not be translated, the readme would be missing those parts. It will probably be outdated soon when there are changes. I would rather have only english readme, than old/inaccurate readme.

alperktt:
> This PR seems to delete the original README?
> 
> But I'm sceptical about adding translated README in general. Unlike the plugin translation where incomplete parts will simply not be translated, the readme would be missing those parts. It will probably be outdated soon when there are changes. I would rather have only english readme, than old/inaccurate readme.

Yeah makes sense :)))

Issue #2012, opened 2025-09-09
Metadata not saved...


Astelteyn:
Added option (default: off) to save generation metadata to images saved directly from the history - not work(((
Option ON. Save from history. No metadata on image(

Acly:
from Discord:
> Acly — 12:33
> which tool are you using to check? did you try others?
> Astelteyn — 12:35
> SD EXIF Editor  and load on civitai. No one see meta... 
> Acly — 12:37
> can you try loading the png back into krita (new document), does it fill out the prompt?
> Astelteyn — 12:40
> yes, it does fill it. So this is what was meant in the update and it is useless for civilai?

So it looks like the metadata is there, only some tools/websites don't pick it up.

@sfisher which reader did you test with? Would you expect Civitai etc. to work with the metadata?

sfisher:
I tested it out in Automatic1111 and Forge under the PNG info tab and it loads.  Sending it to Txt2img or img2img in the program also works and fills the prompt and the information that is present in the file.

It also loads in Irfanview (a windows image viewer).  Press the info button (or press i) and click the "comment" button and it displays the generation information.

It should display in most programs that support PNG iTXt header data.  "iTxt  is used for non-image, descriptive text data — annotations, keywords, comments, copyright, titles, descriptions, and other arbitrary metadata that accompany the image but are not required for decoding or rendering pixels." Many old programs or ones that haven't been updated to use the full set of PNG headers may not display it.  But putting it in PNG headers this way is the way that some standard programs such as Automatic1111 save it.

I understand it will likely display in exiftool, pngcrush, GIMP, ImageMagick identify/convert, or pngtext, though I haven't tested all these.

You can also open in a text editor.  While the binary data of the file will look like garbage you can see the generation metadata toward the top of the file.

Who knows what CivitAI is doing or why they can't parse or decode it since it sounds like the basic information is there.  I haven't tried recently, but in the past I believe they would import some Automatic1111 files and grab the generation data if they were uploaded there.  Though this isn't their primary way of working since they have their own generation infrastructure and save additional data if generating through them, I believe.

There are some possibilities and I don't particularly know what they are doing or why it doesn't import without investigating and spending a bunch more time.

- Possibly they only support tEXt and zTXt chunks but not iTXt chunks.  (Though clearly automatic1111 does support iTxt and reads it ok.)
- Or possibly they expect additional information that isn't supplied in order to parse any of the metadata correctly.
[...]

PR #2128, opened 2025-11-02
Metadata output fix for custom sampler presets

sfisher:
Addresses #2012

Steps and CFG are not set correctly when using a custom setting for sampler with a custom preset.

The user said metadata not showing up when uploaded to CivitAI.  After the user supplied more information and sample file it became obvious that three things were different from the normal saving of metadata which usually works.

- Steps were unknown
- CFG scale was unknown
- The sampler showed additional information besides the sampler name.
  -  It was supplied that way from the Job params rather than these pieces of information coming in under the normal params as `steps` and `guidance`.

After downloading the same model safetensors file and having to set it up with a custom preset for the sampler to use lcm I was able to duplicate the problem.

This PR supplies a fallback for obtaining these settings from the sampler name when they aren't supplied normally in the JobParams if a custom preset is used that isn't built in.

This is a sample upload on civitai after the fix and it fills the steps and cfg to it so it seems to be happier.

Screenshot 2025-11-01 193834

I updated, ran test, ruff, etc.

### DIFF
--- a/ai_diffusion/text.py
+++ b/ai_diffusion/text.py
@@ -1,7 +1,8 @@
 from __future__ import annotations
 import re
 from pathlib import Path
-from typing import Tuple, List, NamedTuple
+from typing import Tuple, List, NamedTuple, Optional
+
 
 from .api import LoraInput
 from .files import FileCollection, FileSource
@@ -230,6 +231,7 @@ def edit_attention(text: str, positive: bool) -> str:
         else f"{open_bracket}{attention_string}:{weight:.1f}{close_bracket}"
     )
 
+
 # in case a custom sampler is used then it may show steps and cfg in there
 def _parse_sampler_embeds(sampler: str) -> Tuple[str, Optional[int], Optional[float]]:
     """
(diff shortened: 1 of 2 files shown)


Acly:
I only guessed that might be the reason, so it's good to have it confirmed.

But I think the change I did (putting steps/guidance in separate metadata fields) should already fix that, and no regex parsing is needed. Did you encounter a scenario where images generated with the current code still produce metadata that does not contain explicit steps/guidance fields?

One way to get old metadata is opening an old document, but IMO it's not worth it to add code to handle that beyond making sure it doesn't crash. It's not that critical, and will solve itself.

sfisher:
Let me double check.  I thought I had pulled all the latest code and used with a new generation and still saw the steps/cfg being tacked onto the checkpoint name without the separate fields.  But I may not have updated correctly or something else.  I'll do another pull and check.

I only saw it happening when a custom setting was used for the checkpoint (such as adding the custom configuration for `lcm` for the checkpoint this user was using).

sfisher:
I tested out the latest main branch and it solves the problem with the custom presets (like below) that put the steps and guidance in the field for the checkpoint name.

Screenshot 2025-11-04 082304

Closing this PR since this seems to be already solved in the latest code.

Verified that the missing steps/CFG were probably what was causing some problems with CivitAI.

PR #2210, opened 2025-12-07
Handful of custom graph specific QoLs

Kebolder:
## Cavnas resize node support

The node currently lives on my node [repo](https://github.com/Kebolder/ComfyUI-Jax-Nodes). I can make a PR for the tooling nodes for the new node code.

* Added support for a new ComfyUI node action: resize_canvas.
* Krita will now listen for ``application/x-krita-command`` messages containing the ``resize_canvas`` action and automatically resizes the active document’s canvas using the provided width and height values.
image

## Custom graph parameter saving across Kra documents

* Added setting section to enable/disable and clear the saved parameter per locally saved workflow.
* Added support for saving and restoring custom workflow parameter values across documents for local workflows.
* Per-workflow global defaults stored in ``settings.custom_params_defaults``, applied automatically when the workflow is used in new documents.
* Limited the parameter saving to saved workflows not remote/temporary ComfyUI web workflows. (These will always use the default parameter values like before)

image

## 2 New buttons under the Custom graph section

* Clear temporary workflows. Useful for when you have a alot of generated workflows whether it be reloading comfy without closing or restarting Krita
* Override saved workflow. As stated, when saving any workflow it will bring up the option to override an existing workflow.
image
image
image

## Updated Custom graph parameter sliders to include input

* Self explanatory 
image

# Why?
I've been using the plugin more and more as of late, slowly shifting from reforge to ComfyUI/Krita but there was some small nitpicks I found a bit annoying about how the plugin felt a bit limited in some regards. So after some research I've conducted some non-invasive updates for users like me who prefer using the Custom graph section.

### DIFF
--- a/ai_diffusion/model.py
+++ b/ai_diffusion/model.py
@@ -816,10 +816,6 @@ class Model(QObject, ObservableProperties):
 
     def set_style(self, style: Style):
         if style is not self._style:
-            if client := self._connection.client_if_connected:
-                styles = filter_supported_styles(Styles.list().filtered(), client)
-                if style not in styles:
-                    return
             if self._style_connection:
                 QObject.disconnect(self._style_connection)
             self._style = style
--- a/ai_diffusion/ui/widget.py
+++ b/ai_diffusion/ui/widget.py
@@ -338,16 +338,14 @@ class StyleSelectWidget(QWidget):
             return
         client = root.connection.client_if_connected
         self._styles = filter_supported_styles(Styles.list().filtered(), client)
+        if self._value not in self._styles:
+            self._styles.insert(0, self._value)
         with SignalBlocker(self._combo):
             self._combo.clear()
             for style in self._styles:
                 icon = theme.checkpoint_icon(resolve_arch(style, client))
                 self._combo.addItem(icon, style.name, style.filename)
-            if self._value in self._styles:
-                self._combo.setCurrentText(self._value.name)
-            elif len(self._styles) > 0:
-                self._value = self._styles[0]
-                self._combo.setCurrentIndex(0)
+            self._combo.setCurrentText(self._value.name)
 
     def change_style(self):
         style = self._styles[self._combo.currentIndex()]
@@ -372,7 +370,10 @@ class StyleSelectWidget(QWidget):
     def value(self, style: Style):
         if style != self._value:
             self._value = style
-            self._combo.setCurrentText(style.name)
+            if style not in self._styles:
+                self.update_styles()
+            else:
+                self._combo.setCurrentText(style.name)
 
 
 class ResizeHandle(QWidget):


Kebolder:
If you have any question, comments, concerns or whatever please feel free to contact me on Discord. I am in the server under "Jax" by my username is Kebolder there!

Acly:
We can merge (some of) these, but I think it would be easier to review and discuss if individual features are separate PRs.

### Canvas resize

Feel free to open a tooling-nodes PR too. Regarding the client code: it would be better to not use `TextOutput` (after all it is not text output). Create something like  a `ResizeCommand` with width/height instead. The server message can be parsed directly in `ComfyClient`, and evaluated in `Model`. It does not have to go to `CustomWorkspace` which shouldn't have anything to do with Krita documents anyway.

### Custom graph parameter saving

It would be nice if it could be automatic without the setting and clear button. For non-custom parameters, stuff like inpaint settings are kept for new documents but eg. prompts are not. For custom parameters it's a bit trickier but still clear that it makes no sense to save layers for example.

### Buttons in custom graph section

Not convinced about those. Clearing abandoned comfyui clients can be improved in various ways, a manual button isn't really a good solution, and it's not such a big problem to require stop-gap.

### Updated Custom graph parameter sliders to include input

No objections, eventually I want to use the Krita sliders, but this is fine in the meantime.

Kebolder:
Apologize on the multiple merge on 1 PR!. I'll keep that in mind and be more mindful on updates next time ^^! 

Fair enough, for the resize section it was the easiest method I could think of at the time since the SendText node existed and worked for me. But I can do a little more direct interactions with the ``ResizeCommand`` no problem.

For the saving. If you'd like I can take a crack at making a entire section dedicated for power user people who like specific things saved in the settings menu. Like I said these where mainly aimed towards the Custom graph section of the plugin, but I'm totally willing to build more on it!

Buttons. I agree now that I think about it a bit more the button for cleaning did seem a bit un-needed and I'll see if I can come up with a more automatic way of keeping it clean.

Acly:
> Apologize on the multiple merge on 1 PR!. I'll keep that in mind and be more mindful on updates next time ^^!

You can still create separate PRs for this, I think it will allow merging faster.

>  If you'd like I can take a crack at making a entire section dedicated for power user people who like specific things saved in the settings menu.

That's not what I meant - I prefer to not add more/any settings at all. They add effort to testing and the more niche they are, the more likely they are to break. The goal should be to pick a reasonable set of parameters that is saved with no user configuratio necessary.

Kebolder:
> You can still create separate PRs for this, I think it will allow merging faster.

Will do, going to clean some things up and replace the textoutput method with a more proper method of rescaling the canvas

> That's not what I meant - I prefer to not add more/any settings at all. They add effort to testing and the more niche they are, the more likely they are to break. The goal should be to pick a reasonable set of parameters that is saved with no user configuratio necessary.

Perhaps I can update the parameter node to have a boolean to tell krita what to save across documents? If not I think anything that uses anything but prompt/text would suffice for custom users.


You or I can close this for now and I'll come back with proper separated PRs and their respective commits. Then we can discuss on those.

PR #2213, opened 2025-12-09
Custom graph resize node support.

Kebolder:
Updated the system to not use the sendtext feature anymore and made a more direct way of telling Krita to resize the canvas to the desired size using the node. 

Also made sure that our upscale image can be previewed first before applying properly.

### DIFF
--- a/ai_diffusion/model.py
+++ b/ai_diffusion/model.py
@@ -816,10 +816,6 @@ class Model(QObject, ObservableProperties):
 
     def set_style(self, style: Style):
         if style is not self._style:
-            if client := self._connection.client_if_connected:
-                styles = filter_supported_styles(Styles.list().filtered(), client)
-                if style not in styles:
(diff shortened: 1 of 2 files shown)


Kebolder:
I realized other commits are leaking into the PR, fixed that and removed those off the main branch.

Acly:
> Also made sure that our upscale image can be previewed first before applying properly.

This made me think, it might be nicer if it wasn't a command that immediately resizes the canvas when the respective node is executed, but rather only attaches the desired size to the output. And it will resize when the result is applied. Resize resolution could be an optional input the the "Krita Output" node, and send alongside the output image message.

Anyway just a thought, I'm also ok with doing it like this.

Kebolder:
That does sound better having a optional input size on the output node, which could help with the "maybe i dont want it this size" before applying that change.

PR #2214, opened 2025-12-09
Override saved custom workflows + Slider input for workflows.

Kebolder:
Reimplemented the slider input feature to this branch.

As for the override I reworked it to be more straight forward, when saving a temporary workflow and naming it the same as a already saved. Krita will pop up asking user to override. If not it will add a suffix like before.


Acly:
This PR still contains the changes for the Resize also, it might lead to conflicts if they're merged like that.

Code looks fine otherwise.

Kebolder:
I'm still learning more about Github itself, not too familiar with how pull request and such. Sorry for any confusion!

Gonna sort this out with branches properly!

PR #2217, opened 2025-12-11
Krita output resize 

Kebolder:
As stated in the #2216 fixed the confusing branches, still wrapping my head around Github entirely with PRs!

Updated the system to use the Krita output node now instead of a entirely different node. Sent a PR on the [Tooling nodes](https://github.com/Acly/comfyui-tooling-nodes) to update the Krita output to include the "Width/Height" which are optional


ai_diffusion/model.py
```diff
@@ -622,6 +633,89 @@ def _finish_job(self, job: Job, event: ClientEvent):
             self.jobs.notify_cancelled(job)
             self.progress = 0
 
+    def _apply_resize_command(self, cmd: ResizeCommand, job: Job):
```
Acly: This could just be
```py
job.params.resize_canvas = cmd.resize_canvas  # or cmd.extent if it stores width/height
```
Please add a `resize_canvas` field to job params, don't use metadata

Since then it's just 1 line probably don't need a function :)

ai_diffusion/model.py
```diff
@@ -622,6 +633,89 @@ def _finish_job(self, job: Job, event: ClientEvent):
...
+            height = int(resize.get("height", 0))
+        except Exception:
+            return None
+
+        if width <= 0 or height <= 0:
+            return None
+
+        return width, height
+
+    def _preview_canvas_size(self, params: JobParams):
```
Acly: I'm not sure it's a good idea to resize on preview. I know you will only get a partial preview otherwise, but for high-res documents with a lot of layers, resize can take several seconds. Waiting seconds when toggling preview sounds annoying. You can always apply and undo, which is maybe better in this case.

ai_diffusion/model.py
```diff
@@ -622,6 +633,89 @@ def _finish_job(self, job: Job, event: ClientEvent):
...
+
+        original = self._preview_canvas_original_extent
+        self._preview_canvas_original_extent = None
+
+        try:
+            self.document.resize_canvas(original.width, original.height)
+        except Exception as e:
+            log.warning(f"Failed to revert canvas size after preview: {e}")
+
+    def _apply_canvas_resize_for_params(self, params: JobParams):
```
Acly: Probably can put this directly into the `apply_result` function. If `resize_canvas` is just a bool it would be something like this:
```py
if params.resize_canvas:
    self.document.resize_canvas(*image.extent)
```

Kebolder:
Updated and remade most of the resizing into the ``params.resize_canvas`` as stated. Was setting up the base lines for a system to allow users to "on the fly" preview the sizes based off the jobs in the history, at least originally that was the idea.

Kebolder:
Would it be better if we do

Run upscale -> Resizes canvas before applying the job or -> Resize Canvas after applying.


I feel like having the resize happen before we apply it is better for the users end since they can see what it would look like if they dont like it just resize it manually?

Acly:
> Run upscale -> Resizes canvas before applying the job or -> Resize Canvas after applying.

Not sure what you mean exactly?

I think resizing for preview is not a good idea, because resize can be slow, and it's destructive (upscale and downscale back to original if you don't like does not restore original image).

Resize on Apply (and possibly undo) is maybe not ideal, but can't think of anything better right now.

ai_diffusion/comfy_client.py
```diff
@@ -927,3 +942,27 @@ def _extract_text_output(job_id: str, msg: dict):
...
+            active = any(bool(item) for item in resize)
+        else:
+            active = bool(resize)
+
+        if not active:
+            return None
+
+        # Use a lightweight dict result; the Krita client will interpret this
+        # as "resize canvas to match image extent" on apply.
+        return ClientMessage(ClientEvent.output, job_id, result={"resize_canvas": True})
```
Acly: Please use the `ResizeCommand` here (with a bool inside). Not sure what a "lightweight dict" is. The code related to previous method via text message can be removed.

Kebolder: Alright, sorry took so long was busy for the last few days. I stripped it down and made it into a pure command for Krita. is this how you'd expect it?

ai_diffusion/model.py
```diff
@@ -622,6 +632,12 @@ def _finish_job(self, job: Job, event: ClientEvent):
             self.jobs.notify_cancelled(job)
             self.progress = 0
 
+    def _apply_resize_command(self, cmd: ResizeCommand, job: Job):
+        try:
+            job.params.resize_canvas = bool(cmd.resize_canvas)
+        except Exception as e:
+            log.warning(f"Failed to store resize command from custom workflow: {e}")
+
```
Acly:
```suggestion
    def _apply_resize_command(self, cmd: ResizeCommand, job: Job):
        job.params.resize_canvas = cmd.resize_canvas

```

I think you can just put this 1 line above, instead of having a function

ai_diffusion/model.py
```diff
@@ -665,6 +681,18 @@ def apply_result(
...
+                extent = image.extent
+                target_width, target_height = extent.width, extent.height
+                current = self.document.extent
+                if current.width != target_width or current.height != target_height:
+                    self.document.resize_canvas(target_width, target_height)
+            except Exception as e:
+                log.warning(f"Failed to resize canvas from custom workflow: {e}")
+            finally:
+                params.resize_canvas = False
+
```
Acly:
```suggestion
        if params.resize_canvas and self.document.extent != image.extent:
            self.document.resize_canvas(*image.extent)
```

ai_diffusion/util.py
```diff
@@ -22,7 +22,11 @@
 
 
 def _get_user_data_dir():
-    if importlib.util.find_spec("krita") is None:
+    spec = importlib.util.find_spec("krita")
```
Acly: This change also doesn't seem related, please revert

Acly (commented):
Yea this looks good, I just added a few more comments to simplify the code

Kebolder:
Will do, some stuff from my experimenting branch leaked over for whatever reason. I'll get back to you once I'm free!

Kebolder:
There we go, reverted the odd changes before and used your suggestions to make the code just straight forward rather then using try/finally. If anything else was weird or got accidently changed on this branch let me know.

PR #2225, opened 2025-12-16
Language update zh-cn, zh-tw

lokpro:
You may hold off if you intend to add additional language items.


Acly:
There are 3 more items now but they are very niche, I think there is no hurry.

PR #2228, opened 2025-12-19
Diffusers local generation backend for CUDA, ROCm, and Apple MPS

bghira:
I'm working on this side project to introduce a diffusers mini server backend into Krita that allows using, for now, Qwen Image Layered, which is a segmentation model that can "infinitely segment" anything.

it has some VRAM use issues and resolution limits, but it's an interesting experimental option, and I wanted to give people a way to try it in Krita.

the Connection system was a bit beyond me, so I implemented a Diffusers specific tab for handling the quantisation options, CPU offload, and the use of RamTorch.

I tested this on an AMD 7900XTX 24G, with int8 precision and RamTorch for CPU offload. It took a minute or so, and produced several layers from an image of a hot air balloon.

to start the thing:

```py
cd ~/.local/share/krita/ai_diffusion/server/diffusers
~/.local/share/krita/ai_diffusion/server/diffusers/venv/bin/python diffusers_server.py --host 127.0.0.1 --port 8189 --device cuda --cpu-offload --vae-tiling --quantization int8 --quantize-transformer --quantize-text-encoder --ramtorch
```

**update, v2**

the diffusers backend is now full-featured and has t2i, i2i, segmentation, t2i+segmentation, and inpainting support.

t2i+segmentation will first generate an image with your selected model and prompt before running the segmentation pipeline on it, with model switchover in between.

Z-Image, Flux, SDXL, and others all work. any model on Huggingface can technically be loaded if Diffusers supports it.

no LoRA or LyCORIS support is planned for now.

**update, v3**

i've fixed some bugs with parameter parsing, the 2-stage workflow for generating an image with eg. Z-Image which then gets immediately segmented now works. also, setting a global quantisation level will no longer override the model-specific quant level.

Z-Image (Fast) can be used to generate image directly for quantised (int8) Qwen Image Layered.

i've noticed that 5 steps for the layering model is way too little with this noise schedule setup, but around 15-20 steps and you can get a dice roll of a residual noise layer that can be deleted to clean the result up.

### DIFF
--- a/ai_diffusion/api.py
+++ b/ai_diffusion/api.py
@@ -199,6 +199,10 @@ class LayeredInput:
     qwen_quantization: str = "none"
     qwen_vae_tiling: bool = True
     qwen_ramtorch: bool = False
+    # Sampler settings (via skrample)
+    sampler: str = "default"  # default, euler, dpm, sdpm, adams, unipc, unip, spc
+    sampler_order: int = 2  # Solver order for higher-order samplers (1-9)
+    schedule: str = "default"  # default, beta, sigmoid, karras
 
 
 @dataclass
@@ -224,6 +228,10 @@ class DiffusersInput:
     quantize_text_encoder: bool = False
     vae_tiling: bool = True
     ramtorch: bool = False
+    # Sampler settings (via skrample)
+    sampler: str = "default"  # default, euler, dpm, sdpm, adams, unipc, unip, spc
+    sampler_order: int = 2  # Solver order for higher-order samplers (1-9)
+    schedule: str = "default"  # default, beta, sigmoid, karras
 
 
 @dataclass
--- a/ai_diffusion/diffusers_client.py
+++ b/ai_diffusion/diffusers_client.py
@@ -390,6 +390,11 @@ class DiffusersClient(Client):
             request["qwen_vae_tiling"] = layered.qwen_vae_tiling
             request["qwen_ramtorch"] = layered.qwen_ramtorch
 
+        # Sampler settings (for both layered_generate and layered_segment)
+        request["sampler"] = layered.sampler
+        request["sampler_order"] = layered.sampler_order
+        request["schedule"] = layered.schedule
+
         # Input image for segmentation mode
         if work.kind == WorkflowKind.layered_segment and work.images:
             if work.images.initial_image:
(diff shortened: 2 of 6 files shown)


Acly:
I wonder what exactly made you create an entire separate server/backend?

I'm not opposed to having another client/server, but full support for all the features (especially the ones not directly related to diffusion) is probably a tall order. And maintaining an extra server only makes sense if it brings proportional benefit over the existing one.

I was sketching qwen-image-layered support yesterday. See #2233 - it only requires few changes and integrates with the existing UI. Still need to add a way to set number of layers somewhere, but apart from that I'm inclined to wait with further integration. Transparency support is very needed, but not convinced this model is the right way.

bghira:
this is about wayyy more than qwen image layered, it's about plugging diffusers in and not having to rely on comfyUI at all. for example flux2 LoRAs don't even load in comfy-core. they made some strange decision to use the flux.1 keys for flux.2. that's neither here nor there - it's a bottleneck to rely on comfyui server only.

to refactor the one server to handle both backends i didn't want to do because i didn't want to have to verify that i didn't break comfyui, as, i don't use it, don't have it installed, and didn't want to set it up (i'm on AMD GPU).

it's done in a way that is going to Just Work for different GPU vendors, whereas Comfy is not.

you don't have to merge it, that's fine.

Acly:
Comfy also works with different GPU vendors by installing the appropriate version of torch, only reason ROCM isn't an option in the automatic installer is that I've no way of testing it. But if this is more of a ideological "Comfy vs Diffusers" thing I won't argue.

> to refactor the one server to handle both backends i didn't want to do because i didn't want to have to verify that i didn't break comfyui

If you want to reuse the UI and all the rest rather than building your own, the idea would be to
1. Implement `Client` (you've already done that and it looks fine from what I can see)
2. Add a branch [in connection.py](https://github.com/Acly/krita-ai-diffusion/blob/0ca8cdc431db43bfd2c2185b253715163963f522/ai_diffusion/connection.py#L96) to create a `DiffusersClient` instead of Comfy/Cloud
3. Maybe add a `ServerMode.diffusers` and UI option in the connection settings to select that instead of Comfy/Cloud

That way it becomes an alternative without interfering with any of the existing server/client implementations.

If you want to easily hack/add additional UI your current approach is probably easier, although I guess overlap with the existing code is tiny at that point and it might as well be a separate project.

PR #2233, opened 2025-12-20
Support Qwen-Image-Layered

Acly:
Works like an Edit model, but decomposes images into transparent layers. Layered images show up as one result in history and expand into layers when applied.

Currently still missing a UI toggle to set number of layers.


Rubenacp:
Qwen Layered stoped working good. Even if I write a promt de result is bad. It generate layers but wrong layers.
CFG: 2.5 - 4.0

image
image

PR #2244, opened 2025-12-23
Update server.py

asmertpc-cloud:
rocm amd gfx110x


asmertpc-cloud:
изображение

изображение

Acly:
So does this only work on gfx110X ?

It would probably be better to use stable/official pytorch index (eg. https://download.pytorch.org/whl/rocm6.4)

(also please merge the other PRs into this one and close them)

Issue #1380, opened 2024-11-11
Feature Request: Make Prompt Parameter in Custom Workflows more usable.


leha-games:
I really like the new custom workflows feature. Amazing work as always!

But prompt parameter node can be made way more usable if it outputs not only text, but also includes loras that in the prompt (it already has loras autocompletion in this field).

I see it like this. There is a new node that has three inputs: model, clip and text append (because we need to append prompts from Krita Style node) and one output something like CONDITIONING.

Also it would be super amazing if it share prompts with other modes like Generate, Live, Animation. Seems like it must throw an error if there are more than one positive or one neagative Prompt node in a workflow.

Thanks!

![image](https://github.com/user-attachments/assets/7d7a9ca7-73c2-4e1b-bbec-aef68311b9db)

RadioVs:
A Style node have prompt outputs which is useless now, it would be good if this node came with a prompt window as in normal mode.
Something like this:
![image](https://github.com/user-attachments/assets/46889a96-228e-4eee-ab9a-4cc01b07cfb9)

leha-games:
> it would be good if this node came with a prompt window as in normal mode. 

Yes, I like the idea

leha-games:
> The main purpose is to give full control & flexibility, so I think the current behavior is fine. It might make sense to offer some high level nodes for convenience. But they'd likely duplicate some functionality that already exists in Comfy or other custom nodes.

Yes, but I really like the convenience of this plugin.

As an example I have a custom workflow where I can upscale image to 4k or maybe 8k with an upscaler and then Tiled Diffusion with control net. It would be useful if prompts is synced with Generate Mode. You can just switch to custom workflow any time and hit generate to get 8k. No need to copy-paste prompts.

And by the way you have custom lora paths searching code. I have to modify [LoRA Tag Loader for ComfyUI ](https://github.com/badjeff/comfyui_lora_tag_loader) extension to get similar behavior of inserting loras from text prompt.

So it would be nice if
1. Syncing prompts between modes 
2. Loading loras from prompt identically as in other modes. 

Thanks. Your work is really top notch.

PR #2276, opened 2026-01-16
Add support for Style & Prompt node in custom workflows

leha-games:
https://github.com/Acly/comfyui-tooling-nodes/pull/57

Closes https://github.com/Acly/krita-ai-diffusion/issues/1380

Features:
- Prompts and Style sync across Generate/Live/Animation and now Graph workspaces
- Full prompt evaluation: wildcards, comments, style merging, LoRA extraction (uses existing implementation, no code duplication)
- LoRAs applied to model output
- Error shown if workflow contains multiple KritaPromptStyle nodes (only one allowed since prompts sync to shared fields)
- "Copy Prompt", "Copy Prompt (Evaluated)", "Copy Style" works as expected across workspaces

### DIFF
--- a/ai_diffusion/custom_workflow.py
+++ b/ai_diffusion/custom_workflow.py
@@ -13,6 +13,7 @@ from PyQt5.QtCore import QMetaObject, QTimer, pyqtSignal
 from .api import WorkflowInput, InpaintContext
 from .client import OutputBatchMode, TextOutput, ClientOutput, JobInfoOutput
 from .comfy_workflow import ComfyWorkflow, ComfyNode
+from .localization import translate as _
 from .connection import Connection, ConnectionState
 from .image import Bounds, Image, Mask
 from .jobs import Job, JobParams, JobQueue, JobKind
@@ -373,6 +374,7 @@ class CustomWorkspace(QObject, ObservableProperties):
     has_result = Property(False)
     outputs = Property({})
     params_ui_height = Property(100, persist=True)
+    validation_error = Property("")
 
     workflow_id_changed = pyqtSignal(str)
     graph_changed = pyqtSignal()
@@ -383,6 +385,7 @@ class CustomWorkspace(QObject, ObservableProperties):
     has_result_changed = pyqtSignal(bool)
     outputs_changed = pyqtSignal(dict)
     params_ui_height_changed = pyqtSignal(int)
+    validation_error_changed = pyqtSignal(str)
     modified = pyqtSignal(QObject, str)
 
     _live_poll_rate = 0.1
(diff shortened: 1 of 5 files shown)


ai_diffusion/custom_workflow.py
```diff
@@ -304,6 +307,19 @@ def workflow_parameters(w: ComfyWorkflow):
             case ("ETN_KritaStyle", _):
                 name = node.input("name", "Style")
                 yield CustomParam(ParamKind.style, name, node.input("sampler_preset", "auto"))
+            case ("ETN_KritaPromptStyle", _):
+                name = node.input("name", "Style")
+                yield CustomParam(ParamKind.style, name, node.input("sampler_preset", "auto"))
+                yield CustomParam(
```
Acly: It's probably enough to just use 1 `yield CustomParam(ParamKind.synced_prompt), name)`

It's only used to detect whether to show the prompt widget, right? The default also makes no sense to me, since the node doesn't have these inputs.

ai_diffusion/custom_workflow.py
```diff
@@ -419,10 +437,21 @@ def _update_workflow(self, idx: QModelIndex, _: QModelIndex):
...
+            self._validate_workflow(self._graph)
             self._metadata = list(workflow_parameters(self._graph))
             self.params = _coerce(self.params, self._metadata)
             self.graph_changed.emit()
 
+    def _validate_workflow(self, wf: ComfyWorkflow):
+        prompt_style_count = sum(1 for _ in wf.find(type="ETN_KritaPromptStyle"))
+        if prompt_style_count > 1:
+            self.validation_error = _(
+                "Workflow contains multiple Krita Prompt Style nodes. "
```
Acly: I'm not 100% sure, but I think you need to put the string in 1 line so it's picked up by the translation script (it's just a crude regex that scans the code for things to translate)

ai_diffusion/model.py
```diff
@@ -517,6 +532,34 @@ async def _generate_custom(self, previous_input: WorkflowInput | None):
...
+            return {}
+        
+        positive = params.get(f"{style_name}/positive_prompt", "")
+        negative = params.get(f"{style_name}/negative_prompt", "")
+        
+        cond = ConditioningInput(positive, negative)
+        arch = resolve_arch(style, self._connection.client_if_connected)
+        prepared = workflow.prepare_prompts(cond, style, seed, arch, FileLibrary.instance())
+        
+        params[f"{style_name}/_prepared"] = {
```
Acly: Rather than putting this fake entry into params, I'd prefer to keep it separate.

Extend `api.py`:
```py
@dataclass
class CustomWorkflowInput:
    workflow: dict
    params: dict[str, Any]
    positive_evaluated: str = ""
    negative_evaluated: str = ""
    loras: list[LoraInput] = field(default_factory=list)
```
that way it's more explicit and typed

Acly (commented):
This looks nice to have! Just had a few change requests.

leha-games:
Thanks for suggestions. Looks more clean now.
Also added proper Style widget "sync" between workspaces

image

leha-games:
Something like this?

Acly:
This should work. There seems to be a conflict now, please rebase onto main so the tests can run.

If you still need to deal with the formatting issues, you can just run `ruff format` and it should pass.

leha-games:
Rebased, formatted, squashed the commits

PR #2284, opened 2026-01-19
Fix missing controlnet range slider for models that support traditional controlnets

nabileisa:
It seems you made an error in 1.46.0 where you are checking for `has_range` in ` models.model_patch.find(self.mode, True)`  instead of  `models.control.find(self.mode, True)` resulting in range slider being hidden in all SD and SDXL models
This fix this issue.


Acly:
Yep, thank you

PR #2358, opened 2026-02-22
Add Illustrious Rectified Flow to support ChenkinNoob Rectified-Flow

FeepingCreature:
Some people over at CivitAI are training a [Rectified Flow version of ChenkinNoob](https://civitai.com/models/2363696). As the NoobXL lineage is generally considered an Illustrious fork, I've added this as an Illu variant. It's basically just "SDXL but with SD3 sampling."

Not sure if this is worth merging, but in case somebody else wants to play with it, I figured I'd put up a PR. It doesn't seem obviously better than VPred Noob in my tests.

edit: I've hardcoded the filename for now. The model identifies as "model_type: eps" so there's actually no good way rn to recognize it as a "SDXL RF" model. Swap the vpred toggle out for a sampling type dropdown? `illu_vpred` is already weird.

### DIFF
--- a/ai_diffusion/client.py
+++ b/ai_diffusion/client.py
@@ -257,7 +257,7 @@ class ClientModels:
         if result := self.resources.get(id.string):
             return result
         # Fallback to epsilon model if v-prediction model not found
-        if id.arch is Arch.illu_v:
+        if id.arch in (Arch.illu_v, Arch.illu_rf):
             if result := self.resources.get(id._replace(arch=Arch.illu).string):
                 return result
         # Search for architecture-agnostic model
--- a/ai_diffusion/control.py
+++ b/ai_diffusion/control.py
@@ -146,7 +146,7 @@ class ControlLayer(QObject, ObservableProperties):
         if client := root.connection.client_if_connected:
             models = client.models.for_arch(self._model.arch)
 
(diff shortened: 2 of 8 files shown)


Acly:
The way it's integrated it will adopt all the control-net & ip-adapter models from Illustrious - does that actually work, are they compatible?

> The model identifies as "model_type: eps" so there's actually no good way rn to recognize it as a "SDXL RF" model. Swap the vpred toggle out for a sampling type dropdown? illu_vpred is already weird.

The v-pred toggle is "legacy" and at this point only needed for some obscure SD1.5 experiments (furry no doubt). Illu v-pred doesn't need this toggle to be on (it's ignored). I don't really want a different sampling toggle either.

As for model arch detection, if the weights are structurally identical to SDXL there's no good way to do it.

---

Since it's not clear yet if this model will be useful, maybe a simple hack to support it would be in `workflow.py` to do
```
if checkpoint.checkpoint.lower() == "chenkinnoobxlv02_v02.safetensors":
    model = w.model_sampling_sd3(model)
```

FeepingCreature:
the controlnets seem to work in testing! they mess things up a bit but not more than they ayways do on noob. it's a straight transfer from a noob finetune so that's expected.

yeah it's a pretty niche ask, I'm entirely fine with waiting if it grows up into something others want to use.

Issue #2374, opened 2026-03-12
Selection feathering is not working with Graph mode


Pauan:
When using Graph mode, create a workflow like this:

<img width="1151" height="374" alt="Image" src="https://github.com/user-attachments/assets/3f3e1537-2912-47ab-bdee-c41a8b53fc99" />

When selecting an area of the image and running that workflow, it gives this result:

<img width="686" height="688" alt="Image" src="https://github.com/user-attachments/assets/7193053d-7bbf-472f-8f8c-6caf53d9c953" />

The selection padding (the black area) is correct, however there is a sharp transition from black to white, even though I have feathering enabled in the settings.

Acly:
Internal workflows use the feather / transition settings to setup denoising and alpha masks, but this depends on a variables and knowledge that don't exist for custom graphs, and is subject to change. I feel that's too much magic to include into custom graphs, if you want feathering in your workflow you can set that up explicitly with the respective nodes.

Automatic crop (with padding) was included because that needs to happen on Krita side and allows transferring smaller image parts.

Pauan:
> if you want feathering in your workflow you can set that up explicitly with the respective nodes.

I tried that already, but there are several problems:

* The feathering is a percentage, not a static pixel value.
* The plugin is using some algorithms to determine the padding / feathering, it's quite difficult to recreate. Especially when the selection is close to the border of the canvas.
* Because the padding is already pre-applied to the mask, using nodes like `Feather Mask` doesn't work, because it will feather the padding instead of the selection.
* The `Krita Selection` node doesn't provide enough information to manually crop the image. It doesn't provide the width / height of the selection, and it doesn't provide any of the options (amount of feathering, amount of padding).

> Automatic crop (with padding) was included because that needs to happen on Krita side and allows transferring smaller image parts.

Yes, and feathering needs to happen on Krita side as well, it cannot be easily done on the ComfyUI side.

If you know of a way to accomplish this within ComfyUI, I would love to hear it.

Acly:
It doesn't happen on Krita side for the built-in functionality either, it's all nodes. You can check them in the dumped workflow. Yes there's a bunch of computation going on to parameterize them, you'd have to do that in the workflow. 

> Because the padding is already pre-applied to the mask, using nodes like Feather Mask doesn't work, because it will feather the padding instead of the selection.

Not sure what node you're using, but that sounds wrong. It should work.

[...]

PR #2401, opened 2026-03-26
Adding in selection bounds to the Krita Selection node

Pauan:
This adds in the selection x / y / width / height bounds to the `Krita Selection` node.

Without this feature, it is impossible for custom workflows to crop / pad / feather selections.

* In `automatic` mode padding is automatically added to the selection, and the custom workflow doesn't know the amount of padding, so it can't crop / feather the selection.

* In `entire image` mode the `offset x` and `offset y` are both 0, and the width / height are the same as the entire canvas, so the custom workflow cannot know the bounds of the selection.

* In `mask bounds` mode the custom workflow has access to the x / y / width / height of the selection.

   But the `Krita Canvas` node is pre-cropped, which makes it impossible to pad the canvas, because padding must happen before cropping.

This PR makes the selection bounds available in all 3 modes.

This makes it possible for custom workflows to calculate the amount of padding in `automatic` mode.

It also makes it possible to do cropping / padding / feathering in `entire image` mode, which is  currently impossible.

Related to https://github.com/Acly/krita-ai-diffusion/issues/2374

### DIFF
--- a/ai_diffusion/api.py
+++ b/ai_diffusion/api.py
@@ -178,6 +178,7 @@ class CustomStyleInput:
 class CustomWorkflowInput:
     workflow: dict
     params: dict[str, Any]
+    selection_bounds: Bounds
     positive_evaluated: str = ""
     negative_evaluated: str = ""
     models: CheckpointInput | None = None
--- a/ai_diffusion/document.py
+++ b/ai_diffusion/document.py
@@ -50,9 +50,12 @@ class Document(QObject):
     def check_color_mode(self) -> tuple[Literal[True], None] | tuple[Literal[False], str]:
         return True, None
 
+    def user_selection_bounds(self) -> Bounds | None:
+        raise NotImplementedError
+
     def create_mask_from_selection(
-        self, mod: SelectionModifiers
-    ) -> tuple[Mask, Bounds] | tuple[None, None]:
+        self, selection_bounds: Bounds | None, mod: SelectionModifiers
+    ) -> Mask | None:
         raise NotImplementedError
 
     def get_image(
@@ -196,20 +199,25 @@ class KritaDocument(Document):
             return False, msg_fmt.format("depth", "8-bit integer", depth)
         return True, None
 
-    def create_mask_from_selection(self, mod: SelectionModifiers):
+    def user_selection_bounds(self):
         user_selection = self._doc.selection()
         if not user_selection:
-            return None, None
+            return None
 
         if _selection_is_entire_document(user_selection, self.extent):
-            return None, None
+            return None
 
-        selection = user_selection.duplicate()
-        original_bounds = Bounds(
-            selection.x(), selection.y(), selection.width(), selection.height()
+        bounds = Bounds(
+            user_selection.x(), user_selection.y(), user_selection.width(), user_selection.height()
         )
-        original_bounds = Bounds.clamp(original_bounds, self.extent)
-        size_factor = original_bounds.extent.diagonal
+        return Bounds.clamp(bounds, self.extent)
+
+    def create_mask_from_selection(self, selection_bounds: Bounds | None, mod: SelectionModifiers):
+        if selection_bounds is None:
+            return None
+
+        selection = self._doc.selection().duplicate()
+        size_factor = selection_bounds.extent.diagonal
         pad_px = max(int(mod.feather_rel * size_factor), mod.feather_min_px)
         pad_px += mod.pad_offset_px
         pad_px += int(mod.pad_rel * size_factor)
@@ -223,7 +231,7 @@ class KritaDocument(Document):
         )
         bounds = Bounds.clamp(bounds, self.extent)
         data = selection.pixelData(*bounds)
-        return Mask(bounds, data), original_bounds
+        return Mask(bounds, data)
 
     def get_image(self, bounds: Bounds | None = None, exclude_layers: list[Layer] | None = None):
         excluded: list[Layer] = []
--- a/ai_diffusion/model.py
+++ b/ai_diffusion/model.py
@@ -514,6 +514,7 @@ class Model(QObject, ObservableProperties):
             is_anim = self.custom.mode is CustomGenerationMode.animation
             seed = self.seed if is_live or self.fixed_seed else workflow.generate_seed()
             canvas_bounds = Bounds(0, 0, *self._doc.extent)
+            select_bounds = None
             bounds = canvas_bounds
             mask = None
 
@@ -521,6 +522,9 @@ class Model(QObject, ObservableProperties):
                 mods = get_selection_modifiers(Arch.sdxl, InpaintMode.fill, self.strength)
                 mask, select_bounds = self._doc.create_mask_from_selection(mods)
                 mask, bounds = self.custom.prepare_mask(selection_node, mask, select_bounds, bounds)
+
+            if select_bounds is None:
+                select_bounds = canvas_bounds
 
             img_input = ImageInput.from_extent(bounds.extent)
             img_input.initial_image = self._get_current_image(bounds, exclude_internal=not is_live)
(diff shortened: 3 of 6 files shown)


Acly:
I still think it might be easier to just use an independent node to get the bounding box, when you need it. Less efficient perhaps, but I doubt it will be noticeable.

Something like this:
```py
           ys, xs = torch.nonzero(mask != 0, as_tuple=True)

           x_min = xs.min().item()
           y_min = ys.min().item()
           x_max = xs.max().item() + 1
           y_max = ys.max().item() + 1

           bbox = (x_min, y_min, x_max, y_max) # or width=x_max-x_min and height=y_max-y_min
           return bbox
```
Would be surprised if this node doesn't exist already somewhere.

Pauan:
> I still think it might be easier to just use an independent node to get the bounding box, when you need it.
>
> Would be surprised if this node doesn't exist already somewhere.

Such a node does not exist. Asking users to implement their own custom Python extension for such a common use case is ridiculous. Most users don't even know Python. And making an extension is not for the faint of heart.

This plugin already perfectly knows the precise bounding box of the selection. It makes perfect logical sense for the `Krita Selection` node to give the selection bounds to the user.

This is not a niche use case. This is an extremely common use case. This is necessary to do *anything* involving selections.

Right now it's impossible to do anything useful with selections. Either the selection is pre-baked and cannot be modified (`automatic` and `mask bounds` mode), or the selection information simply doesn't exist at all (`entire image` mode).

I'm genuinely so confused why you're opposed to this. What's the point of the `Krita Selection` node if it doesn't actually give the user the selection?

Pauan:
I split the functionality into a new `Krita Selection Bounds` node, instead of adding it to the `Krita Selection` node:

krita selection bounds

I think this is slightly cleaner, and it also makes it more clear that the output is always the same regardless of the context.

Acly:
> This is not a niche use case. This is an extremely common use case.

What use case exactly? Not aware of anything that requires a bounding box when you already have a mask. Maybe for cropping, but the Krita selection node already does that for you, and if you want to do it custom way there's nodes that do it with lots of options. If you paint a mask in ComfyUI directly you don't get its bounding box either, doesn't seem to be a big issue.

The most common use for masks in ComfyUI is to use them as denoising mask (Set Latent Noise Mask and related nodes), possibly after processing them a bit (Shrink/Grow/Feather), and for compositing (ImageCompositeMasked and similar). None of these require a bounding box.

Pauan:
> What use case exactly? Not aware of anything that requires a bounding box when you already have a mask.

These are the things you can currently do:

* `automatic` mode:
  * Use automatically calculated padding (calculated in Krita).
  * Use a fixed pixel padding (seems to be kind of broken, but that's a separate issue).
* `mask bounds` mode:
   * Use a fixed pixel padding.
   * Feather on the inside of the selection.

These are the things you CANNOT currently do (but *can* do with my PR):

* `automatic` mode:
   * Feather on the inside of the selection.
   * Use a custom padding that isn't a fixed pixel padding.
* `entire image` mode:
   * Feather on the inside of the selection.
   * Feather on the outside of the selection (VERY useful).
   * Add whatever padding you want (percentage, fixed pixel amount, etc.)
   * Crop using the `Image Crop` and `CropMask` nodes.

      Useful for a wide variety of things... controlnet images, IP adapter images, animations, etc.

> Maybe for cropping, but the Krita selection node already does that for you,

If I use the `automatic` or `mask bounds` node, then it automatically crops the canvas (not necessarily the cropping I want).

But it won't crop other things like `Krita Image Layer`, `Krita Mask Layer`, controlnet images, IP adapter images, other images loaded from disk, or images that I procedurally generate.

I can use `offset x` and `offset y` to crop those things... but I can't do inside feathering. Or outside feathering. Or custom padding.

If I use `entire image` mode then I don't get anything... no cropping, no feathering, no padding. Nothing.

With my PR everything becomes possible. I can now do whatever feathering, padding, and cropping I want.

> and if you want to do it custom way there's nodes that do it with lots of options.

No, there really aren't. I've tried, they don't work properly with the way that the `Krita Selection` node works.

So here is my challenge to you. Create a ComfyUI workflow that:

1. Creates a mask that is the same bounds as the Krita selection (using whatever means you wish).
2. Grows the mask by 32px on each side (using whatever means you wish).
3. Uses the `Feather Mask` node to feather the mask by 32px.

   This means the inside of the Krita selection is pure white, and then it gradually fades to black on the edges outside of the selection.

   This is standard outside feathering, which is very useful.
4. Crops the canvas image (`Krita Canvas` node) to the mask (using whatever means you wish).

This is a very basic task.

This is how I would solve it with my PR:

custom workflow 2

That correctly creates the mask with 32px of feathering outside of the selection, and the canvas correctly cropped to the mask:

outside feathering

> If you paint a mask in ComfyUI directly you don't get its bounding box either, doesn't seem to be a big issue.

This is about selections, not masks. Selections always have a bounding box.

> None of these require a bounding box.

I'm doing significant amounts of padding, feathering, and cropping. I need to do this on MANY different things, like control nets, IP adapter, regions, etc.

I cannot currently do this. But I can easily do this if the selection bounds are provided. This is such basic information that should be provided.

Acly:
> This is about selections, not masks. Selections always have a bounding box.

Maybe this is the misunderstanding? Selections in Krita are masks. Your workflow makes no sense to me because Krita Selection already gives you the mask. Then you additionally want the bounding box of that mask, just to create a mask again?? It seems unnecessarily complex and convoluted.

All padding modes currently allow you to apply feather (inside/outside) to the selection mask. All of them allow you to crop control images (use the selection offset + size of the mask or canvas). You do not need extra outputs.

The only thing you cannot do directly is custom padding, but I think the "automatic" and "mask bounds"+pad are good for 99% of things people use masks for. For the remaining, if you need full flexibility and control, use "entire image" and you can do anything. But you might need some additional nodes, eg. to do custom crops with your own padding you can get the bounds from the mask (with code I pasted above), or use something like [these nodes](https://github.com/lquesada/ComfyUI-Inpaint-CropAndStitch) (there are others too).

---

I will also give a explanation why I don't like this change from software architectural view. I compare it to the alternative solution using a stand-alone node that gets the bounding box from the mask (code in previous comment). 
* This PR is more code and complexity, it touches several places instead of being localized to one implementation
* This PR requires changes in both nodes and plugin creating coupling and dependency
* Stand-alone node is much more powerful and flexible, because it not only works on specific Krita Selection, but an any mask from any source in ComfyUI

So in conclusion, to me, this just seems like more maintenance overhead for something less useful.

Pauan:
> Your workflow makes no sense to me because Krita Selection already gives you the mask.

It doesn't give me the mask that I want. I've given multiple examples of things that `Krita Selection` currently cannot do.

> to do custom crops with your own padding you can get the bounds from the mask (with code I pasted above)

It should not be necessary for the user to create a custom Python extension and node just to do something as simple and basic as this. That's completely ridiculous.

> This PR is more code and complexity, it touches several places instead of being localized to one implementation

The bounds already exists in the plugin code, this is simply exposing it to the user. It is incredibly trivial, it is *not* complex.

That plugin you mentioned is over 1,600 lines of code, and it has 21 parameters. It is absolutely *not* simpler than just providing the selection bounds (which already exist).

That plugin also does not do what I need it to do, even though it has 21 parameters it is lacking features. So it's not a solution.

----

In comparison, this PR is 9 lines of simple code to add the new feature, plus some simple refactoring of `create_mask_from_selection`.

This refactoring is optional, it's not necessary to make the PR work, but is a good idea to do regardless.

In total, with the optional refactoring it is 23 lines of simple code.

> All padding modes currently allow you to apply feather (inside/outside) to the selection mask. All of them allow you to crop control images (use the selection offset + size of the mask or canvas). You do not need extra outputs.

No, it does not. You cannot apply feathering properly, and it cannot do the cropping I need it to do.

*Please* do my challenge that I gave you.

Acly:
Added a simple bounding box from mask node here: https://github.com/Acly/comfyui-inpaint-nodes/commit/dc7ca7515833a41dc83d254d10fbb01599d77071

Closing this, discussion is going nowhere.

Btw here is a workflow for your "basic task". Note that Krita provides the mask and does the crop. Workflow just has to do grow/feather. It's basically 1 node.
image

Pauan:
Nope, that workflow does not work properly:

incorrect mask


Have you considered that maybe this problem is more difficult than you thought?

Perhaps I actually do know what I'm talking about, and I'm not a 10 IQ retard.

PR #2406, opened 2026-03-28
Adding in new Krita Group Layer node

Pauan:
Currently there are three ways to send images to ComfyUI: `Krita Canvas`, `Krita Image Layer`, and `Krita Mask Layer`.

This adds a new `Krita Group Layer` node that sends an entire group of image layers to ComfyUI (along with the masks and names of each image layer).

This is really useful for a variety of situations:

* Sending multiple animation frames outside of `Generate Animation` mode.
* Sending multiple image masks all at once without needing to send them individually.
* Processing an entire batch of images at the same time (upscaling, detailing, etc.)

Originally I wanted to add this feature to the existing `Krita Image Layer` and `Krita Mask Layer` nodes, but that would have needed significant code refactoring. So I made a new node instead.

### DIFF
--- a/ai_diffusion/comfy_workflow.py
+++ b/ai_diffusion/comfy_workflow.py
@@ -1033,6 +1033,9 @@ class ComfyWorkflow:
     def batch_image(self, batch: Output, image: Output):
         return self.add("ImageBatch", 1, image1=batch, image2=image)
 
+    def unbatch_image(self, image: Output):
+        return self.add("RebatchImages", 1, images=image, batch_size=1)
+
     def image_batch_element(self, batch: Output, index: int):
         return self.add("ImageFromBatch", 1, image=batch, batch_index=index, length=1)
 
@@ -1112,6 +1115,10 @@ class ComfyWorkflow:
         image_batch = self.mask_to_image(batch)
         image = self.mask_to_image(mask)
         return self.image_to_mask(self.batch_image(image_batch, image))
+
+    def unbatch_mask(self, mask: Output):
+        image = self.mask_to_image(mask)
+        return self.image_to_mask(self.unbatch_image(image))
 
     def mask_batch_element(self, mask_batch: Output, index: int):
         image_batch = self.mask_to_image(mask_batch)
--- a/ai_diffusion/comfy_workflow.py
+++ b/ai_diffusion/comfy_workflow.py
@@ -1208,6 +1208,9 @@ class ComfyWorkflow:
         return result
 
     def send_list(self, list: list[Input]):
+        if len(list) == 1:
+            return list[0]
+
         output = self.add("ETN_ListEmpty", 1)
 
         for item in list:
--- a/ai_diffusion/comfy_workflow.py
+++ b/ai_diffusion/comfy_workflow.py
@@ -1200,6 +1200,14 @@ class ComfyWorkflow:
         assert result is not None
         return result
 
+    def send_list_str(self, list: list[str]):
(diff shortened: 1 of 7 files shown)


Acly:
> Originally I wanted to add this feature to the existing Krita Image Layer and Krita Mask Layer nodes, but that would have needed significant code refactoring. So I made a new node instead.

I think the following should work:
* add a "group mode" option to Image Layer / Mask Layer nodes which can be "merged" or "batch"
* read the group mode from the plugin and maybe put it into `CustomParam`
* make `collect_parameters` return the projection ("merged") or an image collection of child layers ("batch")
* make the nodes return the image as batch, same as for animation case

Image Layer node should filter image layers and Mask Layer node mask layers, so you might have to use 2 nodes if you have a mixed group. Also no names.

Not sure that fits your use case, but as it is it already feels highly specific to me, and other users are unlikely to discover or understand how to use it with the extra nodes and lists. It would be nice if Comfy had better support for structured data & lists, but as it is, I think keeping it as simple as possible and sticking to batches is better.

Pauan:
> Also no names.

The names are necessary, they are not optional.

For example, if you want to send regions over to ComfyUI, you need to match the region name with the layer. Or if you're sending masks, you need to match the mask name with the layer.

Sending the names opens up a lot of extra possibilities, which is well worth it.

> It would be nice if Comfy had better support for structured data & lists, but as it is, I think keeping it as simple as possible and sticking to batches is better.

ComfyUI's support for data lists is actually pretty robust, and works quite well.

Because the images and names are both a data list, they will be properly zipped together. So everything works as you would expect.

Pauan:
> add a "group mode" option to Image Layer / Mask Layer nodes which can be "merged" or "batch"

Yes, I originally tried that, but it required a lot of refactoring. I can try it again...

Acly:
It should be a small change. I could be wrong, but if it requires lots of refactor / additional code it's probably not worth it.

Pauan:
Yeah it ended up being a pretty small change.

However, adding this feature to `Krita Mask Layer` is more complicated, because it currently only allows you to select transparency / selection layers... not group layers.

So I only made the change to `Krita Image Layer` for now.

ai_diffusion/comfy_workflow.py
```diff
@@ -1200,6 +1207,17 @@ def load_image_and_mask(self, images: Image | ImageCollection):
         assert result is not None
         return result
 
+    def send_list(self, list: list[Input]):
+        if len(list) == 1:
+            return list[0]
+
+        output = self.add("ETN_ListEmpty", 1)
```
Acly, line 1214: Why does this use custom list nodes?

Can't ComfyUI built-in nodes be used (`PrimitiveString`, `CreateList`)?

ai_diffusion/workflow.py
```diff
@@ -1458,9 +1459,14 @@ def get_param(node: ComfyNode, expected_type: type | tuple[type, type] | None =
             case "ETN_Parameter":
                 outputs[node.output(0)] = get_param(node)
             case "ETN_KritaImageLayer":
-                img, mask = w.load_image_and_mask(get_param(node, (Image, ImageCollection)))
+                layer: CustomLayerInput = get_param(node, CustomLayerInput)
+                img, mask = w.load_image_and_mask(layer.images)
+                if not layer.is_batch:
```
Acly, line 1464: I think it would be better to always return a batch. The node output type can stay the same and it's less confusing for flatten/animation. Rebatch can be added in the workflow if a list is needed.

PR #2412, opened 2026-04-02
Security: Auto-update accepts unsigned metadata and artifacts

tuanaiseo:
## Problem

The updater trusts `version`, `url`, and `sha256` from `/plugin/latest`, then verifies the downloaded ZIP against that same server-provided hash. If the update API/CDN or transport endpoint is compromised, an attacker can deliver a malicious plugin package with a matching hash, resulting in code execution when the plugin is loaded.

**Severity**: `high`
**File**: `ai_diffusion/updates.py`

## Solution

Use a detached signature scheme (for example Ed25519) for update metadata/artifacts, pin the public verification key in the client, and reject updates that fail signature verification.

## Changes

- `ai_diffusion/updates.py` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/ai_diffusion/updates.py
+++ b/ai_diffusion/updates.py
@@ -1,6 +1,7 @@
 import hashlib
 import os
 import shutil
+from base64 import b64decode
 from enum import Enum
 from pathlib import Path
 from tempfile import TemporaryDirectory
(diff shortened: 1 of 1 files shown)


Acly:
I don't think `cryptography` package is available in Krita's embedded python?

PR #2421, opened 2026-04-03
Adding in option to disable double click on previews

### LINKED ISSUE #2293: Option to disable double click
I often toggle the preview on and off by clicking on the generated image, so I can compare the generated image to the canvas.

But if I click too fast it counts as a double click and applies it, which I don't want.

I want to disable the double click behavior completely, so the only way to apply the generated image is to click the `Apply` button.

Pauan:
Prevents accidentally applying an image.

Defaults to true, so this doesn't change existing behavior.

Fixes #2293


ai_diffusion/ui/custom_workflow.py
```diff
@@ -965,8 +965,9 @@ def _splitter_moved(self, pos: int, index: int):
         self.model.custom.params_ui_height = self._splitter.sizes()[0]
 
     def apply_result(self, item: QListWidgetItem):
-        job_id, index = self._history.item_info(item)
-        self.model.apply_generated_result(job_id, index)
+        if settings.preview_double_click_apply:
```
Acly, line 968: Haven't tested it, but this looks like it would prevent applying an image completely, not just via double click, but also the apply button

Instead the conditional should be at the point where the handler is registered, ie. in `HistoryWidget.__init__` where it says `self.itemDoubleClicked`

Acly:
We can keep the setting, but I'd prefer no UI for now. I don't think it's super vital and the interface settings easily become cluttered. Remove the translation strings and UI modification.

PR #2422, opened 2026-04-03
Increased seed input range (0 to 2^32-1)

LennyArchive:
Very small workaround made to allow the user to have the same range of seeds as A1111 WebUi, making it easier to recreate images generated on different tools.


Acly:
Thanks!

PR #2439, opened 2026-04-10
Security: Potential arbitrary file read via unsanitized language identifier

tomaioo:
## Summary

Security: Potential arbitrary file read via unsanitized language identifier

## Problem

**Severity**: `Low` | **File**: `ai_diffusion/localization.py:L72`

`language` is read from a settings file and directly interpolated into a filename (`f"{language}.json"`) under the language directory. Values like `../../somefile` may traverse directories and cause unintended file reads if the target is valid JSON with expected keys.

## Solution

Validate `language` against a strict allowlist/pattern (e.g., `^[a-z]{2}([_-][A-Za-z0-9]+)?$`) and/or resolve the final path and enforce it stays within the language directory before opening.

## Changes

- `ai_diffusion/localization.py` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/ai_diffusion/localization.py
+++ b/ai_diffusion/localization.py
@@ -1,4 +1,5 @@
 import json
+import re
 from pathlib import Path
 from typing import NamedTuple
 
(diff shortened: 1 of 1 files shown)


FeepingCreature:
I mean... it's the settings file. in what scenario can you access the settings file but not just any other file you want?

PR #2442, opened 2026-04-11
<feature request> automatically change float docker to a window

zerobikappa:
https://github.com/user-attachments/assets/55bd7b8f-3425-477e-b1e3-260b4c44a859

## Why is necessary
The docker takes too much space. Even though user can make the docker float, the docker always keep above on the main window.
After change the float docker to a window, user can easily use system shortcuts (Win+Up, Win+Down, Alt+Tab, etc...) to minimize/maximize/switch the windows. It is useful to users who only have one small monitor.

## How to change the docker into a window
(1) Double click titlebar of the docker.
(2) Click float button of the docker.
(3) Drag the docker to non-docker-area then release.

## How to change the window back to docker
Just simply close the extension window (Alt+F4, or click the close button), then it will change back to docker.

## How to move the docker to other position
(If it is under window mode, change back to docker mode first.) Drag the docker to other docker-area then release.

### DIFF
--- a/ai_diffusion/extension.py
+++ b/ai_diffusion/extension.py
@@ -49,8 +49,8 @@ class AIToolsExtension(Extension):
         notifier = Krita.instance().notifier()
         notifier.setActive(True)
         notifier.applicationClosing.connect(self.shutdown)  # type: ignore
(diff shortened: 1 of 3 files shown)


zerobikappa:
So sorry that I forgot to create a new branch. Shall I close this PR and create a new PR with a different branch?

zerobikappa:
create another PR with a new branch name.

PR #2450, opened 2026-04-14
Add Anima architecture support

FeepingCreature:
[Anima](https://huggingface.co/circlestone-labs/Anima/) is a promising up and comer! Not super slow like Flux, but also with much better prompt understanding than anything OpenCLIP based, it's still in testing but it looks to be the first model I'd cautiously say is a straight improvement over SDXL.

ComfyUI supports it ootb, so there's almost nothing to do. There are only two finicky details: there's no model class for it yet so we have to recognize by filename, and the preview releases don't bake in VAE and TE (because they're frozen) so they have to be loaded separately even tho it's a standard safetensors load node.

Also of course there's no controlnets yet, though a first derived model [already popped up](https://civitai.com/models/2385278/animayume?modelVersionId=2851312), so it's probably just a matter of time.

(thanks claude for code!)

### DIFF
--- a/ai_diffusion/resources.py
+++ b/ai_diffusion/resources.py
@@ -95,6 +95,7 @@ class Arch(Enum):
     qwen_e = "Qwen Edit"
     qwen_e_p = "Qwen Edit Plus"
     qwen_l = "Qwen Layered"
+    anima = "Anima"
     zimage = "Z-Image"
 
     auto = "Automatic"
@@ -134,6 +135,8 @@ class Arch(Enum):
             return Arch.qwen_l
         if string == "qwen-image":
             return Arch.qwen
+        if string == "anima" or (string == "unknown" and "anima" in filename):
+            return Arch.anima
         if string in {"z-image", "zimage"}:
             return Arch.zimage
         return None
@@ -187,7 +190,7 @@ class Arch(Enum):
 
     @property
     def supports_attention_guidance(self):
-        return self in [Arch.sd15, Arch.sdxl, Arch.illu, Arch.illu_v]
+        return self in [Arch.sd15, Arch.sdxl, Arch.illu, Arch.illu_v, Arch.anima]
 
     @property
     def supports_cfg(self):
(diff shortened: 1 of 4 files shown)


ai_diffusion/resources.py
```diff
@@ -134,6 +135,8 @@ def from_string(string: str, model_type: str = "eps", filename: str | None = Non
             return Arch.qwen_l
         if string == "qwen-image":
             return Arch.qwen
+        if string == "anima" or (string == "unknown" and "anima" in filename):
```
Acly, line 138: I think this should already be detected, but probably as `lumina2`, can you check?
Since base Lumina 2 isn't supported, we might as well map it to Anima.

ai_diffusion/workflow.py
```diff
@@ -119,6 +119,8 @@ def load_checkpoint_with_lora(w: ComfyWorkflow, checkpoint: CheckpointInput, mod
     match (model_info.format, model_info.quantization):
         case (FileFormat.checkpoint, Quantization.none):
             model, clip, vae = w.load_checkpoint(model_info.filename)
+            if arch is Arch.anima:
```
Acly: This shouldn't be needed if the model is correctly located in `diffusion_models` / `unet` folder.

Technically people could make a Anime "checkpoint" with integrated TE/VAE and this would prevent it from being loaded... not sure how important that is, but since other models don't do this atm I'd remove it for now.

ai_diffusion/resources.py
FeepingCreature: Not that I can tell? `{"anima-preview.safetensors": {"base_model": "unknown", "is_inpaint": false}, "anima-preview2.safetensors": {"base_model": "unknown", "is_inpaint": false}, "anima-preview3-base.safetensors": {"base_model": "unknown", "is_inpaint": false}, "animayumeTunedV04.RBPP.safetensors": {"base_model": "unknown", "is_inpaint": false}}` that's with current ComfyUI/ETN

ai_diffusion/workflow.py
FeepingCreature: ah- you're right! I had them in the wrong folder, thanks.

ai_diffusion/resources.py
Acly: Ah okay, looks like it has its own architecture now in ComfyUI.

Could you add it to external-tooling-nodes? I think it might be very simple, same as https://github.com/Acly/comfyui-tooling-nodes/pull/63 basically, just map `Anima` -> `anima`

Acly (approved):

PR #2451, opened 2026-04-14
<feature request> automatically change float docker to a window

zerobikappa:
https://github.com/user-attachments/assets/55bd7b8f-3425-477e-b1e3-260b4c44a859

## Why is necessary
The docker takes too much space. Even though user can make the docker float, the docker always keep above on the main window.
After change the float docker to a window, user can easily use system shortcuts (Win+Up, Win+Down, Alt+Tab, etc...) to minimize/maximize/switch the windows. It is useful to users who only have one small monitor.

## How to change the docker into a window
(1) Double click titlebar of the docker.
(2) Click float button of the docker.
(3) Drag the docker to non-docker-area then release.

## How to change the window back to docker
Just simply close the extension window (Alt+F4, or click the close button), then it will change back to docker.

## How to move the docker to other position
(If it is under window mode, change back to docker mode first.) Drag the docker to other docker-area then release.


Acly:
This seems like overkill to me. 

You can already arrange the docker using Krita's docker system. If this is only about using Alt+Tab and similar to toggle the docker on/off, that could be solved much more simply by having a configurable shortcut inside Krita that toggles the docker.

zerobikappa:
> This seems like overkill to me.
> 
> You can already arrange the docker using Krita's docker system. If this is only about using Alt+Tab and similar to toggle the docker on/off, that could be solved much more simply by having a configurable shortcut inside Krita that toggles the docker.

(1) It is too difficult to resize docker:

https://github.com/user-attachments/assets/4f9cf0b5-066b-4b16-8040-c1259ea4e78e

https://github.com/user-attachments/assets/3f2a6aac-dd08-45da-b623-1aa68e41e663

(2) Sometime I would like to minimize krita main window to do other things but keep the plugin on foreground to monitor the progress (because it takes long time for my GPU to generate images).

https://github.com/user-attachments/assets/8c5d1643-e7a8-41a9-959b-c9d41ce998bd

https://github.com/user-attachments/assets/e256fea9-f5c2-478a-9cfc-47761e0bfd9a

(3) Just my opinion, it is not worth remembering an extra shortcut which is only for plugin on/off. I would prefer to use desktop environment's shortcuts, which are most used in my daily work.


I have many other reasons for this, but in short words, an individual window is more flexible/user-friendly than a float-docker in most scenarios.

Acly:
I'm not quite convinced, sorry. I can see you have your reasons, but the existing docker and split window behaviors of Krita work okay for the most part, and adding this amount of extra code to do it in a slightly different way is not worth it IMO. Especially since it isn't really discovarable, so I don't expect anyone else to pick it up.

I don't see myself using it either as it breaks basic assumptions about windows and alt+tab behavior (almost always one per app or document, never seen separate windows for widgets).

Makes most sense to keep it in your fork I think.

Issue #2467, opened 2026-04-27
Image upload fails ("Unable to write") when server aborts PUT request on cache hit


rbeurskens:
When an image is used multiple times, can can be cached by ConfyUI.
In my case, this caused Krita to show an "Unable to write" error when that happened.
This was because on the PUT request, the server returned a cache-hit json, and then closed the connection server side for me:
```Powershell
curl.exe -v -X PUT --data-binary "@some-image.jpg" http://<Server Host>:8188/api/etn/image/xxxxxxxx
*   Trying <Server Host>:8188...
* Connected to <Server Host> (<IP>) port 8188
* using HTTP/1.x
> PUT /api/etn/image/xxxxxxxx HTTP/1.1
> Host: <Server IP>:8188
> User-Agent: curl/8.13.0
> Accept: */*
> Content-Length: <image-length>
> Content-Type: application/x-www-form-urlencoded
> Expect: 100-continue
>
< HTTP/1.1 100 Continue
<
< HTTP/1.1 200 OK
< Content-Type: application/json; charset=utf-8
< Content-Length: 20
< Date: Sun, 26 Apr 2026 02:48:41 GMT
< Server: Python/3.12 aiohttp/3.13.5
<
{"status": "cached"}* Send failure: Connection was aborted
```
I solved this by changing the code to first send a HEAD request and only send to PUT if it resulted in a not found response code.
Fixes in #2468 worked for me, so why not share it ? :)

rbeurskens:
Here's a log sample, for a bit more context: 
```
RuntimeError: Error uploading input image to ComfyUI: Unable to write
2026-04-24 11:32:05,121 ERROR Error uploading image a5206f1b: Unable to write
2026-04-24 11:32:05,121 ERROR Unhandled exception while processing Job[id=3409332e-f111-423e-a3c9-fbdd1c4920a1]
Traceback (most recent call last):
  File "%AppData%\Roaming\krita\pykrita\ai_diffusion\comfy_client.py", line 552, in upload_images
    await self._put(f"api/etn/image/{id}", data)
  File "%AppData%\Roaming\krita\pykrita\ai_diffusion\comfy_client.py", line 280, in _put
    return await self._requests.put(f"{self.url}/{op}", data)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ai_diffusion.network.NetworkError: Unable to write

[...]

PR #2468, opened 2026-04-27
Enhancement: Check ComfyUI server cache before image upload

rbeurskens:
Should solve #2467 
Also saves some network traffic by ~not~ only sending the payload if ~HEAD~ PUT with `Expect: 100-Continue` returns ~200~ 100 . ~(in contrast to PUT)~


Acly:
I've never encountered the "Unable to write" error before, does that happen all the time for you or with particular image size? Which OS?

I don't think an additional HEAD request is a necessarily good trade of, now every upload needs an extra round trip, and the not-cached case is probably far more common. 

I consider sending some of the upload data even if it's cached okay, but it would require a way to selectively ignore this particular error, which might be difficult if Qt doesn't return the actual response and 200 status code. Since I can't reproduce it, it's hard to tell.

Cleaner solution might be to start the PUT with a `Expect: 100-continue` header which is less overhead than the HEAD request, but I haven't tested it yet.

rbeurskens:
For me, consistent from consecutive prompts using one or more of the same source image(s) when using a custom ComfyUI server (latest) running on Kubuntu LTS 24.04 on a 2012 era PC (MSI Z77 GD65 mainboard, i7 3770K, 16GB). Krita running on Windows 10 12gen i7,16GB.
I will test your suggestion and let you know if that fixes it too.
The issue could be caused by the combination of the modern Linux running on 2012 hardware (in particular when it comes to power management of the onboard ethernet adapter)

Images were not really large (this one was ~1MB )

Acly:
Might just have to do with latency, I mostly test local with Krita & Comfy running on the same machine.

Note that the `100-continue` solution would require support by the server too (ie. in comfyui-tooling-nodes)

rbeurskens:
Yes, probably. I never had this when running the local managed server. After all, the loopback likely doesn't have this issue. But with an actual remote it's different. (maybe even worse with a cloud server instead of one in the LAN)
I have the latest `comfyUI-tooling-nodes` node cloned on the server. As far as I know the header is a high level protocol feature that does not need hardware/driver level support to work, as long as the HTTP listener supports it. (`aiohttp`)

The problem is that, as you also said, `QtReply` will only process response data if it sent the full payload successfully. No matter if the `PUT` is sent with `Expect-100` or without, if the server sees it has the image cached before the client has finished sending the payload, it will send the `200 OK` response with the `{ "status" : "cached" }` json and close the connection immediately. This will cause `QNetworkReply::RemoteHostClosedError`.
If the lightweight HEAD it too much, we would need to handle the request on a lower level and handle the request manually using `QTcpSocket`, and keep monitoring server response for `200 OK` and stop once it's found to prevent the error `QNetworkAccessManager` would trigger.

Acly:
Not sure why you think it needs to be handled at a lower level, the `100-Continue` mechanism is just HTTP.

As far as I understand you need to change the client to send `Expect-100` in header, change the server to respond with `100-Continue` if uncached or `200` if cached, and the client to send the data when it receives `100` response.

I'm not 100% sure it can be done with `QNetworkManager`, if it's not possible then I'd prefer to just silence the "Unable to write" error.

rbeurskens:
Because `100 Continue` does not solve the issue. (`curl` already sends this header under the hood, and as you may see in the output in my example, it still results in a disconnect before the client can send the payload.) In the high level method with `QNetworkReply`, I tried hooking in a handle error method and handle any `QNetworkReply::RemoteHostClosedError`, but the socket is aborted by the server and it is a known behavior if the response does not return with a 'success', `QNetworkReply` does not offer a way to read any response headers or data (it's like an 'all-or-nothing' black box.) , so I can't check the `content-length`, `content-type` and the content itself to verify if despite the disconnect the server's response is complete in order to know if we are dealing with a real error or a cache hit.
And if we need to rely on altering the server (in this case `ConfyUI` and we may not always have access to the server backend) in its response behavior, that doesn't sound like a very flexible solution.

rbeurskens:
A bit more context of what I tried before: I added an error handler in `network.py` in the `http()` method:
```python
        def debug_log(err):
            status = reply.attribute(QNetworkRequest.HttpStatusCodeAttribute)
            log.info(f"DEBUG: Error {err} occurred. Status: {status}")

        reply.errorOccurred.connect(debug_log)
```
This resulted in the following: (note status is `None`, not `100` or `200`)
Error 99 is `QNetworkReply.UnknownNetworkError`
```
2026-04-26 08:06:56,581 INFO DEBUG: Error 99 occurred. Status: None
2026-04-26 08:06:56,612 ERROR Error uploading image 1d59b868: Unable to write
2026-04-26 08:06:56,612 ERROR Unhandled exception while processing Job[id=6e7d7e08-55a2-4e7e-970f-78fb0828dc58]
Traceback (most recent call last):
  File "%AppData%\Roaming\krita\pykrita\ai_diffusion\comfy_client.py", line 555, in upload_images
    await self._put(f"api/etn/image/{id}", data2, timeout=120)
  File "%AppData%\Roaming\krita\pykrita\ai_diffusion\comfy_client.py", line 280, in _put
    return await self._requests.put(f"{self.url}/{op}", data, timeout=timeout)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ai_diffusion.network.NetworkError: Unable to write
```

rbeurskens:
It CAN be handled on a lower level with the `Qt` library, but needs more manual handling, like this (I still need to test this and HTTPS support can still be added if needed - draft by Copilot):
```python
    def put_socket(
        self,
        url: str,
        data: QByteArray | bytes,
        timeout: float | None = None,
        bearer: str | None = None,
    ):
        self._cleanup()
        if isinstance(data, bytes):
            data = QByteArray(data)
        assert isinstance(data, QByteArray)

        request = _SocketPutRequest(self, url, data, timeout, bearer)
        self._socket_puts.add(request)
        return request.future
        
class _SocketPutRequest:
    def __init__(
        self,
        manager: "RequestManager",
        url: str,
        data: QByteArray,
        timeout: float | None,
        bearer: str | None,
    ):
        self._manager = manager
        self._url = QUrl(url)
        self._data = bytes(data)
        self._timeout = timeout
        self._bearer = bearer
        self._future = asyncio.get_running_loop().create_future()

        self._socket = QTcpSocket()
        self._buffer = bytearray()
        self._headers_sent = False
        self._waiting_for_continue = False
        self._final_status: int | None = None
        self._final_headers: dict[str, str] = {}
        self._final_body = bytearray()
        self._final_content_length: int | None = None
        self._upload_started = False
        self._upload_offset = 0

        self._connect_signals()
        self._connect_socket()

    @property
    def future(self) -> asyncio.Future:
        return self._future

    def _connect_signals(self):
        self._socket.connected.connect(self._on_connected)
        self._socket.readyRead.connect(self._on_ready_read)
        self._socket.bytesWritten.connect(self._on_bytes_written)
        self._socket.errorOccurred.connect(self._on_socket_error)
        self._socket.disconnected.connect(self._on_disconnected)

    def _connect_socket(self):
        if self._url.scheme().lower() == "https":
            self._fail(
                NetworkError(
                    int(QNetworkReply.NetworkError.ProtocolFailure),
                    "Socket PUT does not support HTTPS",
                    self._url.toString(),
                )
            )
            return

        host = self._url.host()
        port = self._url.port(80)
        self._socket.connectToHost(host, port)

    def _on_connected(self):
        if self._future.done():
            return

        path = self._url.path() or "/"
        if self._url.hasQuery():
            path += f"?{self._url.query()}"

        host = self._url.host()
        port = self._url.port(80)
        host_header = f"{host}:{port}" if port != 80 else host

        headers: list[tuple[str, str]] = [
            ("Host", host_header),
            ("User-Agent", "krita-ai-diffusion"),
            ("Accept", "*/*"),
            ("Connection", "close"),
            ("Content-Type", "application/octet-stream"),
            ("Content-Length", str(len(self._data))),
            ("Expect", "100-continue"),
        ]

        headers.extend(self._manager.socket_headers(self._bearer))

        request = [f"PUT {path} HTTP/1.1"]
        request.extend([f"{k}: {v}" for k, v in headers])
        request_bytes = "\r\n".join(request).encode("utf-8") + b"\r\n\r\n"

        self._socket.write(request_bytes)
        self._headers_sent = True
        self._waiting_for_continue = True

        # Some servers ignore Expect and wait for payload immediately.
        asyncio.get_running_loop().call_later(0.25, self._ensure_upload_started)
        if self._timeout is not None:
            asyncio.get_running_loop().call_later(self._timeout, self._on_timeout)

    def _on_timeout(self):
        if self._future.done():
            return
        self._socket.abort()
        self._fail(
            NetworkError(
                int(QNetworkReply.NetworkError.TimeoutError),
                "Connection timed out, the server took too long to respond",
                self._url.toString(),
            )
        )

    def _ensure_upload_started(self):
        if self._future.done() or self._upload_started or self._final_status is not None:
            return
        self._start_upload()

    def _start_upload(self):
        if self._future.done() or self._upload_started or self._final_status is not None:
            return
        self._upload_started = True
        self._waiting_for_continue = False
        self._pump_upload()

    def _on_bytes_written(self, _count: int):
        if self._future.done():
            return
        self._pump_upload()

    def _pump_upload(self):
        if self._future.done() or self._final_status is not None:
            return

        # Keep outgoing buffer bounded so incoming data can be processed quickly.
        max_inflight = 256 * 1024
        chunk_size = 64 * 1024

        while self._socket.bytesToWrite() < max_inflight and self._upload_offset < len(self._data):
            end = min(self._upload_offset + chunk_size, len(self._data))
            chunk = self._data[self._upload_offset : end]
            self._upload_offset = end
            self._socket.write(chunk)

    def _on_ready_read(self):
        if self._future.done():
            return

        self._buffer.extend(bytes(self._socket.readAll()))
        try:
            self._process_http_stream()
        except Exception as e:
            self._fail(e)

    def _process_http_stream(self):
        while True:
            if self._final_status is None:
                parsed = self._try_parse_response_headers()
                if parsed is None:
                    return

                status, headers = parsed
                if status == 100:
                    self._waiting_for_continue = False
                    self._start_upload()
                    continue

                self._final_status = status
                self._final_headers = headers
                content_length = headers.get("content-length")
                if content_length is not None:
                    try:
                        self._final_content_length = int(content_length)
                    except ValueError:
                        self._final_content_length = None

            if self._final_content_length is None:
                # Body length unknown; wait for disconnect to complete.
                self._final_body.extend(self._buffer)
                self._buffer.clear()
                return

            if len(self._buffer) < self._final_content_length:
                return

            self._final_body.extend(self._buffer[: self._final_content_length])
            del self._buffer[: self._final_content_length]
            self._complete_from_final_response()
            return

    def _try_parse_response_headers(self) -> tuple[int, dict[str, str]] | None:
        marker = self._buffer.find(b"\r\n\r\n")
        if marker < 0:
            return None

        raw_header = bytes(self._buffer[:marker])
        del self._buffer[: marker + 4]
        lines = raw_header.decode("latin-1", errors="replace").split("\r\n")
        if not lines or not lines[0].startswith("HTTP/"):
            raise RuntimeError(f"Invalid HTTP response from {self._url.toString()}")

        status_parts = lines[0].split(" ", 2)
        try:
            status = int(status_parts[1])
        except Exception as e:
            raise RuntimeError(f"Invalid HTTP status from {self._url.toString()}") from e

        headers: dict[str, str] = {}
        for line in lines[1:]:
            if not line or ":" not in line:
                continue
            key, value = line.split(":", 1)
            headers[key.strip().lower()] = value.strip()
        return status, headers

    def _on_disconnected(self):
        if self._future.done():
            return

        if self._final_status is not None:
            self._final_body.extend(self._buffer)
            self._buffer.clear()
            self._complete_from_final_response()
            return

        self._fail(
            NetworkError(
                int(QNetworkReply.NetworkError.RemoteHostClosedError),
                "Connection closed before response was received",
                self._url.toString(),
            )
        )

    def _on_socket_error(self, _error: QTcpSocket.SocketError):
        if self._future.done():
            return

        # If final HTTP response was already seen, a disconnect error is expected.
        if self._final_status is not None:
            return

        self._fail(
            NetworkError(
                int(QNetworkReply.NetworkError.NetworkSessionFailedError),
                self._socket.errorString(),
                self._url.toString(),
            )
        )

    def _complete_from_final_response(self):
        if self._future.done() or self._final_status is None:
            return

        status = self._final_status
        data = bytes(self._final_body)
        if 200 <= status < 300:
            content_type = self._final_headers.get("content-type", "")
            if "application/json" in content_type:
                self._future.set_result(json.loads(data or b"{}"))
            else:
                self._future.set_result(data)
            self._socket.abort()
            self._manager.release_socket_put(self)
            return

        msg = data.decode("utf-8", errors="replace") if data else f"HTTP {status}"
        self._fail(
            NetworkError(
                int(QNetworkReply.NetworkError.UnknownServerError),
                msg,
                self._url.toString(),
                status,
            )
        )

    def _fail(self, error: Exception):
        if not self._future.done():
            self._future.set_exception(error)
        self._socket.abort()
        self._manager.release_socket_put(self)
```

Acly:
Yes it absolutely needs server side support! My hope was it can just be added here: https://github.com/Acly/comfyui-tooling-nodes/blob/main/api.py#L351-L366

This code is maintained by me so we can change it without modifying ComfyUI.

rbeurskens:
Ah, cool. That is good to know. ~~Not sure if it would help dealing with how the disconnect is handled, as I think this is likely a problem on the underlying netwerk/adapter/driver, but~~ it can open more options to implement a solution.

We could modify it something like this: (untested draft)
```python
    @_server.routes.put("/api/etn/image/{id}")
    async def put_image(request: web.Request):
        try:
            id = request.match_info.get("id", "")
            if id in image_cache:
                # If client sent Expect: 100-continue, this 200 OK 
                # tells it: "Don't send the data."
                # request.release() ensures any stray bytes are handled.
                await request.release() 
                return web.json_response(dict(status="cached"), status=200)

            # Accessing request.content now triggers the '100 Continue' 
            # signal to the client to start uploading.
            content_type = request.headers.get("Content-Type", "application/octet-stream")
            data = bytearray()
            async for chunk, _ in request.content.iter_chunks():
                data.extend(chunk)

            image_cache.insert(id, bytes(data), content_type)
            return web.json_response(dict(status="success"), status=201)
        except Exception as e:
            # If upload fails mid-way, request.release() helps clean up
            await request.release()
            return web.json_response(dict(error=str(e)), status=500)
```
.. and modify the client code to make the decision based on the `accept-100` to proceed uploading as you suggested earlier:
```python
    def http(
        self,
        method: str,
        url: str,
        data: dict | QByteArray | None = None,
        timeout: float | None = None,
        bearer: str | None = None,
    ):
        self._cleanup()

        request = self._prepare_request(url, timeout, bearer)

        assert method in ["GET", "POST", "PUT", "HEAD"]
        if method == "POST":
            # (...)
        elif method == "PUT":
            if isinstance(data, bytes):
                data = QByteArray(data)
            assert isinstance(data, QByteArray)
            
            # --- Handshake Optimization ---
            # Tell the server we have a payload but want to check headers first.
            # This prevents Error 99/Socket Error 10 by allowing the server 
            # to return "cached" before the client starts the upload.
            request.setRawHeader(b"Expect", b"100-continue")
            
            request.setHeader(
                QNetworkRequest.KnownHeaders.ContentTypeHeader, "application/octet-stream"
            )
            request.setHeader(QNetworkRequest.KnownHeaders.ContentLengthHeader, data.size())
            reply = self._net.put(request, data)
        else:
            reply = self._net.get(request)
     
        future = asyncio.get_running_loop().create_future()
        self._requests[reply] = Request(url, future)
        return future
```

Acly:
This is what I mean: https://github.com/Acly/comfyui-tooling-nodes/pull/65

I don't have time to fully test right now, maybe you can check if it works. I can review/merge later.

rbeurskens:
Sure I can test it as I have the environment to reproduce the problem. I will let you know.

Acly:
I merged the server-side changes. This should already help, but I'm not sure it uses the efficient path.

Did you ever check if the client sends the `Expect 100` header already, or  if it's easy to make it do that (without low-level changes)?

rbeurskens:
> I merged the server-side changes. This should already help, but I'm not sure it uses the efficient path.
> 
> Did you ever check if the client sends the `Expect 100` header already, or if it's easy to make it do that (without low-level changes)?

I tested it with repeated uploads with curl.  And this already looks like how it should look :
```powershell
PS>> curl.exe -v -X PUT --data-binary "@my_image.jpg" http://<myserver>:8188/api/etn/image/1d59b868
*   Trying <myserver>:8188...
* Connected to <myserver> (<server_ip>) port 8188
* using HTTP/1.x
> PUT /api/etn/image/1d59b868 HTTP/1.1
> Host: <myserver>:8188
> User-Agent: curl/8.13.0
> Accept: */*
> Content-Length: 1067265
> Content-Type: application/x-www-form-urlencoded
> Expect: 100-continue
>
* Done waiting for 100-continue
* upload completely sent off: 1067265 bytes
< HTTP/1.1 201 Created
< Content-Type: application/json; charset=utf-8
< Content-Length: 21
< Date: Thu, 14 May 2026 01:09:47 GMT
< Server: Python/3.12 aiohttp/3.13.5
<
{"status": "success"}* Connection #0 to host <myserver> left intact
```
```powershell
PS>> curl.exe -v -X PUT --data-binary "@my_image.jpg" http://<myserver>:8188/api/etn/image/1d59b868
*   Trying <myserver>:8188...
* Connected to <myserver> (<server_ip>) port 8188
* using HTTP/1.x
> PUT /api/etn/image/1d59b868 HTTP/1.1
> Host: <myserver>:8188
> User-Agent: curl/8.13.0
> Accept: */*
> Content-Length: 1067265
> Content-Type: application/x-www-form-urlencoded
> Expect: 100-continue
>
< HTTP/1.1 200 OK
< Content-Type: application/json; charset=utf-8
< Content-Length: 20
< Date: Thu, 14 May 2026 01:10:42 GMT
< Server: Python/3.12 aiohttp/3.13.5
<
{"status": "cached"}* Connection #0 to host <myserver> left intact
```
So server side should be good. I'll modify the client code in this PR to handle it like `curl` does. Indeed no low level code should be needed indeed, I expect. Likely a very small change.

rbeurskens:
I have tested it with the small adjustment I thought that would work (just adding the header), but it seems `Qt`~5~ (5 and 6) has no support for this workflow, and simply ignores it. (if I use a `QtBuffer` instead of a `bytearray`, which should allow it to wait for the server response, the entire process crashes with `QtNetworkManager`.)
From what I found this is a `Qt`~5~ 5 and 6 issue, and since we are stuck to it because Krita uses it (~not `Qt6`, where it might be fixed~ EDIT: it is not), we unfortunately still need a `QTcpSocket` approach when sending this header to properly handle this (default) http workflow.

Acly:
Nevermind then, if it's not possible with Qt's HTTP features I don't think we want it. Reimplementing this with sockets is not worth the complexity.

Qt6 migration is imminent, not sure if it will help but worth checking out.

rbeurskens:
I'm implementing it for myself anyway, and it will be easy to toggle it only for this workflow. ~I'll share it when finished. I'm already testing it.~ I pushed the implementation in the PR. (there's some `log.debug` calls in it, I can remove them if desired.) 
I checked out `Qt6`, and it does not support the `100-Continue` handshake in `QtNetworkManager` either. It would be the choice of the socket implementation to implement and use the `100-Continue` handshake, or if that is considered too complex, I would say, we can use the `HEAD` approach as alternative. It's not a single request like PUT/Expect-100, but still pretty lightweight and not as complex.
Ignoring/suppressing the error is no (clean) option to me, and also hides real errors.

Silly-Jarv:
These code changes solved the problem for me.

I use a GPU solution via runpod.io


Edit: My solution up until now has always been to change the color of a few pixels in the image. That allowed me to get through one generation before the error occurred again.

rbeurskens:
> These code changes solved the problem for me.
> 
> I use a GPU solution via runpod.io
> 
> Edit: My solution up until now has always been to change the color of a few pixels in the image. That allowed me to get through one generation before the error occurred again.

I'm happy to hear it was helpful for you too and the workaround to bypass the cache is no longer needed. Glad I shared it.

PR #2479, opened 2026-05-07
fix: canvas race

tekakutli:
Hello @acly, thank you for your efforts.

I had been suffering from a recursion bug that prevented me from opening a Krita file that had a custom workflow run on it, so I went ahead and tried to fix it. This patch works, at least for me. I couldn't find a way to shrink the patch any further. Whether you find it acceptable or not, it at the very least points to the problem and a solution. So far, I haven't encountered any further errors.

Problem:
- Krita AppImage with Python 3.13 causes RecursionError in pathlib._local.py when `Path.is_relative_to()` is called on style filepaths
- Plugin attempts to access canvas/layers before Krita UI is fully initialized, causing viewport corruption and freezing

Root causes:
1. Python 3.13 pathlib has a bug where `is_relative_to()` can enter infinite recursion on certain Path objects
2. `LayerSelect._update()` and `ImageDiffusionWidget.update_content()` execute during Krita startup, before the canvas is ready

Solution:
- Replace `is_relative_to()` with safe string comparison in `Style.filename`
- Add delayed initialization (QTimer.singleShot) to prevent early canvas access
- Wrap `deepcopy` in try-except with JSON fallback for workflow graphs

Files changed:
- style.py: replace is_relative_to() with string comparison
- comfy_workflow.py: safe deepcopy with RecursionError fallback
- ui/custom_workflow.py: delay LayerSelect initialization
- ui/diffusion.py: delay ImageDiffusionWidget activation

Fixes: RecursionError, canvas viewport corruption, Krita freeze on startup

### DIFF
--- a/ai_diffusion/comfy_workflow.py
+++ b/ai_diffusion/comfy_workflow.py
@@ -79,7 +79,13 @@ class ComfyWorkflow:
         # Re-order nodes such that a nodes inputs are always processed before the node itself.
         while queue:
             id = queue.pop(0)
-            node = deepcopy(existing[id])
+            # Safe deepcopy to avoid recursion on circular references
+            try:
+                node = deepcopy(existing[id])
+            except RecursionError:
+                # Fallback: manual copy for recursive structures
+                import json
+                node = json.loads(json.dumps(existing[id]))
             class_type = node.get("class_type")
             if class_type is None:
                 log.warning(f"Workflow import: Node {id} is not installed, aborting.")
--- a/ai_diffusion/ui/custom_workflow.py
+++ b/ai_diffusion/ui/custom_workflow.py
@@ -63,37 +63,60 @@ class LayerSelect(QComboBox):
         super().__init__(parent)
         self.param = None
         self.filter = filter
+        self._initialized = False
 
         self.setContentsMargins(0, 0, 0, 0)
         self.setMinimumContentsLength(20)
         self.setSizeAdjustPolicy(QComboBox.SizeAdjustPolicy.AdjustToMinimumContentsLength)
         self.currentIndexChanged.connect(lambda _: self.value_changed.emit())
 
-        self._update()
-        root.active_model.layers.changed.connect(self._update)
+        # Don't access root.active_model here - it causes recursion during model loading
+        # Use a delayed initialization instead
+        from PyQt5.QtCore import QTimer
+        QTimer.singleShot(500, self._delayed_init)
+
+    def _delayed_init(self):
+        self._initialized = True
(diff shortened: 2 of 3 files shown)


Acly:
Could you share the workflow that triggered the `deepcopy` recursion?

> Python 3.13 pathlib has a bug where is_relative_to() can enter infinite recursion on certain Path objects

What bug is that? Do you have any more info on this?

---

The document-not-ready-yet issues need a better solution, probably somewhere close to `canvasChanged`. Previous plugin versions always had a built-in delay there, but since Krita 5.3 this sometimes lead to the widget not updating content after opening documents. I plan to revisit that in next few days.

tekakutli:
It's precisely when I installed Krita 5.3.1 that my issues started, although it had been a while since I had both updated and used ComfyUI, so I can't point at exactly when.

It was a stretch to blame a bug, and it was my mistake not to investigate the claim further. The claim is the output of a prompt in which I asked for a summary of the steps I had so far done in the patch. I'm actually using Python 3.14.4.

And while I started with it, it seems like changes to style.py were unnecessary after fixing the actual issues, so I just removed that part. I changed ui/diffusion.py and force-pushed, and it seems like it still werks.

I have not perceived the widget not updating content using this patch.

Here is one of the workflows; any of them impeded me from opening the Krita files:
https://gist.github.com/tekakutli/1186be8df419e8bd865508a082636a83

Acly:
Possible alternative fix: https://github.com/Acly/krita-ai-diffusion/commit/0cfad1ec9f1ee37d7cb214553f77dc9d7b084fa6

@tekakutli could you check if this works for you? I wasn't really able to reproduce any recursion issues or noticeable errors at all, but it might be a timing issue. There were some errors in the log and layer assignments in the UI changed, those are gone now

tekakutli:
indeed, 0cfad1e fixed it for me, thank you a lot Acly

Acly:
uhh, interesting. Is that related to the changes, or it was happening before?

I opened a bunch of files and switched between them during testing yesterday and didn't encounter a crash so far

Acly:
ah nvm I respond to deleted comment I think :)

tekakutli:
I deleted it because I wanted to test it properly, but indeed, right now it only works for the first AI-diffusion-tainted document. I can't figure out what exactly is triggering the 'Segmentation fault (core dumped) krita' error other than the 'opening up a second ai-diffusion-tainted file'

Acly:
Recursion can sometimes result in a segfault. It might still be related to your custom workflow somehow? Does it happen without custom workflows?

PR #2491, opened 2026-05-16
Krita 6.x / Qt6

Acly:
* Migrate to PyQt6 for Krita 6.x
* Use Krita's unified slider-spinbox widgets


FeepingCreature:
Testing! All seems to p much look the same. The qt6 application window doesn't always autoresize properly, and it seems I can't ctrl-enter in the prompt box anymore to submit, and I miss the strength text box emotionally, but it all seems pretty much as before.

There seems to be extra space between the rows of render tiles with image size 1200x800.
Screenshot_20260531_153650

edit: The step count when scrolling the strength slider doesn't seem to be going smoothly as it did on the text box.

Acly:
@FeepingCreature 

Did Ctrl+Enter ever work? Shift+Enter works (maybe not as intuitive, but Ctrl+Enter is already very global in Krita)

> The step count when scrolling the strength slider doesn't seem to be going smoothly as it did on the text box.

Mouse wheel scroll or dragging?

No idea where this thumbnail padding  comes from...

FeepingCreature:
Retesting.

I can get used to shift-enter! That works fine, thanks.

We still have the weird overhang. It's like there's an invisible button there. edit: Only happens if w > h afaict. edit edit: no! also with h > w!

Screenshot_20260628_223518

My problem with the scrollbar is that the mousewheel goes in 5-percent jumps regardless of stepcount, when really I'm usually aiming for precise step numbers. I'm getting used to clicking exactly where I want to go though.

polytect:
Can't wait for Both Tools to work using Krita 6.0

PR #2524, opened 2026-06-12
Add Anima regional control support

Sen-sou:
Added Anima regional control support. Regional control requires an optional (https://huggingface.co/Sen-sou/Anima-LLLite-Regional-Controlnet) model to work best. Since Anima-LLLite models are already supported in krita, therefore added a new controlnet type 'regional'. It doesnt require explicit control image, it automatically combines the current regions to form a control image. Regions can still work without the control model but the region effect is relatively weaker.

Edit: Both 'regional' control net and region attention should be able to work independently. But currently region attention works independently but 'regional' controlnet needs regions to work.

### DIFF
--- a/ai_diffusion/model/control.py
+++ b/ai_diffusion/model/control.py
@@ -124,10 +124,10 @@ class ControlLayer(QObject, ObservableProperties):
             bounds = None  # ignore mask bounds, use layer bounds
 
         image = None
-        if self.mode is ControlMode.regional:
-            image = self._regional_control_image(bounds, time)
+        if self._is_anima_segmentation:
+            image = self._segmentation_control_image(bounds, time)
             if image is None:
-                image = self._regional_control_image_from_layer(layer, bounds, time)
+                image = self._segmentation_control_image_from_layer(layer, bounds, time)
         if image is None:
             image = layer.get_pixels(bounds, time)
 
@@ -145,7 +145,11 @@ class ControlLayer(QObject, ObservableProperties):
         strength = self.strength / self.strength_multiplier
         return ControlInput(self.mode, image, strength, (self.start, self.end))
 
-    def _regional_control_image_from_layer(
+    @property
+    def _is_anima_segmentation(self):
+        return self.mode is ControlMode.segmentation and self._model.arch is Arch.anima
+
+    def _segmentation_control_image_from_layer(
         self, layer: Layer, bounds: Bounds | None, time: int | None
     ):
         bounds = bounds or Bounds.from_extent(self._model.document.extent)
@@ -153,7 +157,7 @@ class ControlLayer(QObject, ObservableProperties):
         image.draw_image(layer.get_pixels(bounds, time))
         return image
 
-    def _regional_control_image(self, bounds: Bounds | None, time: int | None):
+    def _segmentation_control_image(self, bounds: Bounds | None, time: int | None):
         from .region import RegionLink
 
         bounds = bounds or Bounds.from_extent(self._model.document.extent)
@@ -175,6 +179,8 @@ class ControlLayer(QObject, ObservableProperties):
         return image if has_region_layer else None
 
     def generate(self):
+        if not self.can_generate:
+            return
         self._generate_job = self._model.generate_control_layer(self)
         self.has_active_job = True
 
@@ -195,10 +201,7 @@ class ControlLayer(QObject, ObservableProperties):
                     self.error_text = _("The server is missing the ClipVision model") + f" {search}"
                     is_supported = False
 
-            if self.mode is ControlMode.regional and models.arch is not Arch.anima:
-                self.error_text = _("Not supported for") + f" {models.arch.value}"
-                is_supported = False
-            elif self.mode.is_ip_adapter and models.arch.supports_edit:
+            if self.mode.is_ip_adapter and models.arch.supports_edit:
                 is_supported = True  # Reference images are merged into the conditioning context
             elif self.mode.is_ip_adapter and models.ip_adapter.find(self.mode) is None:
                 search_path = resources.search_path(ResourceKind.ip_adapter, models.arch, self.mode)
@@ -238,7 +241,7 @@ class ControlLayer(QObject, ObservableProperties):
                 is_supported = False
 
         self.is_supported = is_supported
-        self.can_generate = is_supported and self.mode.has_preprocessor
+        self.can_generate = is_supported and self.mode.has_preprocessor and not self._is_anima_segmentation
 
     def _update_is_pose_vector(self):
         self.is_pose_vector = self.mode is ControlMode.pose and self.layer.type is LayerType.vector
@@ -443,7 +446,6 @@ control_mode_text = {
     ControlMode.blur: _("Unblur"),
     ControlMode.stencil: _("Stencil"),
     ControlMode.hands: _("Hands"),
-    ControlMode.regional: _("Regional"),
 }
 
 
--- a/ai_diffusion/model/control.py
+++ b/ai_diffusion/model/control.py
@@ -5,12 +5,13 @@ from pathlib import Path
 from typing import Any, NamedTuple
 
 from PyQt5.QtCore import QObject, Qt, QUuid, pyqtSignal
+from PyQt5.QtGui import QColor
 
 from .. import util
 from ..backend import resources
 from ..backend.api import ControlInput
 from ..backend.resources import Arch, ControlMode, ResourceKind, resource_id
-from ..image import Bounds, Extent, Image
+from ..image import BlendMode, Bounds, Extent, Image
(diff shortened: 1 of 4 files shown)


Acly:
I don't think this is a good way to add it - Regions are already the mechanism to define regional prompts, adding another way that is somehow related but needs to be setup manually on top is not intuitive (and also a hassle).

I will look at the attention couple PR in tooling nodes and hopefully get around to testing it soon. Don't have access to my GPU at home atm so might take a bit. Definitely want to try and get that in and see how it performs.

If the control lllite is needed on top, it should be implemented as part of using regions, without any further setup or knowledge required on user side. But it will increase code complexity quite a bit, and without being able to actually assign prompts to colors in the control image it seems a bit random. I'm not really sure what the control model actually learned here...? Would be interested what your validation looked like.

Sen-sou:
The controlnet model originally started as an experiment for training a regional control model, thinking that the prompts would associate with their respective colored region in the control image, but that didnt happen, so it ended up with just characters which follow those colored region without adhering to specific character prompts. That was the regional controlnet.
And you are already aware how weak attention masking is for Anima model, so i thought pairing the two together would cancel out each others drawbacks (attention couple masks text tokens and regional controlnet masks the image tokens), and it does remarkably well. Together they produce strong regional adherence compared to their individual results.
But to use it in krita, would require both to be active at the same time. Attention couple/masking was can be easily merged into whats already there, as for having the lllite patch run along with the regions would make the implementation complicated, so i just added it as a regional controlnet type.

Acly:
Attention masking is always somewhat "weak" when it conflicts with what the model thinks it needs to do to keep the image consistent. But IMO it's better than to force regions too hard and sacrifice composition. Slight prompt drift is much easier to fix in follow-up steps (eg. "Refine Region") than bad overall composition.

I'm not entirely convinced on the controlnet - for relatively simple setups that are already decent without it, it improves adherence, but also seems to fit content to the mask shape on a structural level somewhat (not necessarily desired). And it changes the image style completely.
For challenging setups where Anima already struggles it doesn't help that much, and composition is noticeable worse.

Apart from complexity I'm concerned that the small scale of the CN made it learn to fit the dataset and not just the general task. I'm not even entirely sure what the task was supposed to be, since the CN doesn't interact with prompts - it seems to go in the direction of segmentation CN.

Sen-sou:
Those are fair concerns.

The regional controlnet model's original goal is to learn true prompt to region assignment, but it didnt end up behaving that way. From my testing, it behaved closer to a spatial/layout conditioning model that encourages subjects to occupy the specified regions. Thats why i found it worked best when paired with attention masking (which i tested in comfyui). The control model influences where the content appears and attention masking influences which prompt affected the region. Both approaches have limitation, but together they produced better adherence in my testing.
Regarding changing image style, maybe i didn't test or notice it much, but it might be possible, since the dataset was only 580 images. The project was largely experimental and it showed promising results and a lot of people seems to have really positive reviews on this. I also have plans to increase the dataset to 4000+ images for the next version, so that might fix the image style issue. As for the content structure fitting to mask shape, lower strength can help reduce that (but in my opinion, to able to define your own shape for the content gives much more control to the user).

Here are some example and tests from other users:
https://x.com/ushifuto/status/2067186826981634372
https://x.com/sndiws184/status/2066718730999263500
https://x.com/kiyoshi_shin/status/2066339654731960403
https://x.com/kurayamimousou/status/2066211793899557218
https://x.com/uni_nori_/status/2066164887236284542
https://x.com/uni_nori_/status/2066049172168642784
https://x.com/higana_yogana/status/2065654661261246785

As for integration, i understand your concern about adding it as separate regional workflows. But from my testing, and you know too, that attention masking by itself doesn't perform well at all, and it literally doesnt work with complex regions. For that reason, i view this controlnet model as a complementary requirement for regions to properly work in krita. 

Implementation-wise i will leave it to your discretion.

Acly:
Defining shape isn't something the region feature ever did or is supposed to do though. Usually it's combined with a sketch or depth image for structural guidance.

It sounds like the region CN fills a similar purpose, and is quite close the "Segmentation" control net. I would suggest to just expose it as Anima's segmentation control layer for now, since it doesn't have one currently.

If you want to generate a segmentation guidance image from the defined regions, this could be added as a button next to the "Generate control image" button. 

If I got it correctly, the CN expect regions as plain color blobs in the image. Current `_regional_control_image` implementation just uses the layer content, which often will not be plain color, ie. when editing an existing image. Should probably change that to just take coverage/alpha from the layer, and fill it with distinct plain color.

Sen-sou:
That does sound better. I'll take up your suggestions.

Edit: I think all the changes can be kept, and 'regional' just need to be changed to 'segmentation'

Acly:
My idea was to add a new button next to the existing generate/pre-processor button, which generates a segmentation image from current regions. It should work the same as the pre-processors, ie. create an explicit new layer with the colored regions that becomes the source for the control layer. Also should be available to all models, not a special Anima thing.

The behavior as it is in this PR isn't intuitive at all, Anima has special behavior that just ignores any configured layer and generates something else behind the scenes which users never see and somehow depends on regions when there is no obvious connection.

Also the function that generations the segmentation image needs to draw plain colors, NOT the region layer content.

Sen-sou:
> My idea was to add a new button next to the existing generate/pre-processor button, which generates a segmentation image from current regions. It should work the same as the pre-processors, ie. create an explicit new layer with the colored regions that becomes the source for the control layer. Also should be available to all models, not a special Anima thing.
> 
> The behavior as it is in this PR isn't intuitive at all, Anima has special behavior that just ignores any configured layer and generates something else behind the scenes which users never see and somehow depends on regions when there is no obvious connection.
> 
> Also the function that generations the segmentation image needs to draw plain colors, NOT the region layer content.

totally forgot about that. Made the changes, please check

ai_diffusion/model/control.py
```diff
@@ -139,7 +157,63 @@ def to_api(self, bounds: Bounds | None = None, time: int | None = None):
...
+    def _segmentation_region_image(
+        self, layer: Layer, bounds: Bounds, color: tuple[int, int, int]
+    ):
+        mask = layer.get_mask(bounds)
+        image = Image.create(bounds.extent, fill=QColor(*color, 255))
+        image._qimage.setAlphaChannel(mask._qimage)
+        return image
+
     def generate(self):
+        if not self.can_generate:
```
Acly: any particular reason this was added?

ai_diffusion/model/control.py
```diff
@@ -139,7 +157,63 @@ def to_api(self, bounds: Bounds | None = None, time: int | None = None):
...
+        image = Image.create(bounds.extent, fill=Qt.GlobalColor.white)
+        has_region_layer = False
+        root = self._model.active_regions
+
+        layers = [
+            layer
+            for layer in root.layers.all
+            if root.find_linked(layer, RegionLink.direct) is not None
+            and layer.compute_bounds().area > 0
+            and Bounds.intersection(bounds, layer.bounds).area > 0
```
Acly:
```suggestion
            and Bounds.intersection(bounds, layer.compute_bounds()).area > 0
```

ai_diffusion/model/control.py
```diff
@@ -139,7 +157,63 @@ def to_api(self, bounds: Bounds | None = None, time: int | None = None):
         strength = self.strength / self.strength_multiplier
         return ControlInput(self.mode, image, strength, (self.start, self.end))
 
+    def generate_segmentation(self):
+        if self.mode is not ControlMode.segmentation:
+            return
+
```
Acly:
```suggestion
        assert self.mode is ControlMode.segmentation
```

Acly (commented):
Looks good, just some small stuff and the linting error

ai_diffusion/model/control.py
Sen-sou: safeguard for unintended calls

ai_diffusion/model/control.py
Sen-sou: changed this, but assertion can be disabled right?

ai_diffusion/model/control.py
Acly: what unintended calls? if there is no way to hit this code please remove it.

ai_diffusion/model/control.py
Acly: technically yes, but they are not disabled in Krita (and I'm not sure what your point is?)

Acly (approved):
LGTM, Lint check still needs to be fixed, but I probably won't be able to merge until weekend anyway

ai_diffusion/model/control.py
Sen-sou: yea sorry about that, i didnt check there were only two calls for this, and they would be hidden. Removed it.

PR #2528, opened 2026-06-14
Show newest generation results at the top of the history

carlogerli00-collab:
## What

The generation history now inserts newly finished results at the **top** of the list — the prompt header goes to row 0 and the new thumbnails right below it — instead of appending them at the bottom. Older results move down. Prompt grouping is preserved (same-params jobs still share a single header), and the view auto-scrolls to the top (instead of the bottom) when the user is already there.

## Why

The Generate/Refine buttons sit at the top of the panel, so having fresh results appear right under them is much more convenient than scrolling to the bottom of a long history after every generation. This matches the common "newest first" feed behaviour.

## How

In `HistoryWidget` (`ai_diffusion/ui/generation.py`):
- `add()` inserts the group header at row 0 and the result thumbnails right below it (`insertItem`) instead of `addItem`; it scrolls to top (not bottom) when the user is already at the top.
- `_add_item()` gained an optional `row` argument to support insertion at a position.
- `rebuild()` scrolls to top.

No new settings/dependencies; `ruff format`/`ruff check` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/ai_diffusion/ui/generation.py
+++ b/ai_diffusion/ui/generation.py
@@ -155,7 +155,7 @@ class HistoryWidget(QListWidget):
             return  # Only finished diffusion/animation jobs have images to show
 
         scrollbar = self.verticalScrollBar()
-        scroll_to_bottom = scrollbar and scrollbar.value() >= scrollbar.maximum() - 4
+        scroll_to_top = scrollbar is not None and scrollbar.value() <= 4
 
         if not JobParams.equal_ignore_seed(self._last_job_params, job.params):
             self._last_job_params = job.params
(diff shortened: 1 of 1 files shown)


Acly:
Insert at bottom is intentional, thumbnails moving when results come in asynchronously is very irritating.

carlogerli00-collab:
> Insert at bottom is intentional, thumbnails moving when results come in asynchronously is very irritating.

I think it's much better, also images at the bottom often result in being cut out and bugged.. but do as you wish, I'm customizing it on my end anyway

PR #2529, opened 2026-06-14
Preserve style architecture in custom workflows

mtsmfm:
## Summary

When using Graph mode with Anima, workflows that include `ETN_KritaStyle` can fail with:

```text
Server error: Text Encoder 'clip_l' for SD 1.5 not found
```

image

image

image

The style preset may explicitly use the Anima diffusion architecture, but custom workflow parameter collection was passing style.get_models(...) directly into CustomStyleInput. In some cases, especially with
generic checkpoint filenames, that leaves the collected style models at the default SD 1.5 architecture instead of the architecture resolved from the connected model metadata.

As a result, an Anima checkpoint can be routed through the SD 1.5 text encoder path.

This change resolves the style architecture using the available checkpoint metadata before creating CustomStyleInput, so custom workflows preserve the intended architecture such as Anima.

### DIFF
--- a/ai_diffusion/model/custom_workflow.py
+++ b/ai_diffusion/model/custom_workflow.py
@@ -22,7 +22,14 @@ from PyQt5.QtCore import (
 
 from .. import eventloop
 from ..backend.api import CustomStyleInput, InpaintContext, WorkflowInput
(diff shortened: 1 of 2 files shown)


Acly (approved):

Acly:
Saw some people hitting this problem, thanks for the fix!

PR #2530, opened 2026-06-14
Fix plugin auto-update install layout

tarelion:
## Summary
- Treat older/equal plugin versions from the update API as up-to-date instead of available updates.
- Install release archives using the actual Krita resource layout: copy `ai_diffusion/` into the plugin directory, `ai_diffusion.desktop` into `pykrita/`, and `ai_diffusion.action` into `actions/`.
- Clean up the nested `ai_diffusion/` folder and misplaced `ai_diffusion.desktop` created by the previous updater path.
- Add local regression tests that do not require the cloud update service.

## Motivation
The current updater copies the extracted archive root directly into the installed plugin directory. Release archives contain `ai_diffusion/` and `ai_diffusion.desktop` at their root, so this creates a nested `ai_diffusion/ai_diffusion/` installation and leaves the active plugin files unchanged.

It also treats any version mismatch as an available update. If the update API reports an older compatible version than the installed plugin, the UI keeps showing the update screen and can repeatedly attempt a downgrade.

## Test Plan
- `pytest tests/test_updates.py::test_update_check_ignores_older_versions tests/test_updates.py::test_update_installs_release_package_layout -q`
- `pytest tests/test_updates.py -q`
- `ruff format --check ai_diffusion/model/updates.py tests/test_updates.py`
- `ruff check ai_diffusion/model/updates.py tests/test_updates.py`
- `pyright ai_diffusion/model/updates.py tests/test_updates.py`

I also ran `pytest tests --ci -q`; unrelated environment/setup checks failed locally because `tests/server` is not installed and PyQt5's bundled Qt was built against OpenSSL 1.x while the runtime is OpenSSL 3.x. The new updater tests pass independently.

### DIFF
--- a/ai_diffusion/model/updates.py
+++ b/ai_diffusion/model/updates.py
@@ -2,6 +2,7 @@ import hashlib
 import os
 import shutil
 from enum import Enum
+from itertools import zip_longest
 from pathlib import Path
 from tempfile import TemporaryDirectory
 from typing import NamedTuple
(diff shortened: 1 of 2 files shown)


Acly:
Layout issue should be fixed with https://github.com/Acly/krita-ai-diffusion/commit/05285083bc2460e205f703a9bd03623dc363e50a (is just 1 line)

Not sure about the more complex version comparison, there usually isn't a situation where the latest version on the server is older (except during development, but auto-updates don't make sense there anyway)

PR #2539, opened 2026-06-22
Iterate seed after apply

wraybowling:
This PR is my attempt to add a feature for a quirk that bothered me. In live mode, I prefer punching in 1 as my seed number and iterating by 1 as I accept changes. This allows me to return to seeds that worked well just by using small simple numbers that are easy to remember.

At the time of this PR being created I have not added translations yet and I am still testing. Big fan of the project! First time contributing here. Not vibe coded, fwiw.

### DIFF
--- a/ai_diffusion/backend/workflow.py
+++ b/ai_diffusion/backend/workflow.py
@@ -61,6 +61,8 @@ def generate_seed():
     # Currently only using 32 bit because Qt widgets don't support int64
     return random.randint(0, 2**32 - 1)
 
+def iterate_seed(old_seed):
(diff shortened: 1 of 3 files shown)


Acly:
Let's call it "Increment" instead of iterate.

What happens when loading settings which use the old true/false? Might have to add some kind of migration.

Also the comparisons to the enum name don't look like they're actually valid code...

wraybowling:
Noted! I see I also need to learn how to run your tests to get a passing build. I'll have another pass at it soon. Thanks for looking

PR #2613, opened 2026-09-21
Support Qwen Image 2.1 (t2i and edit)

fukc-gihtub:
Tested in Krita 6.0.4

Caveat: the model supports transparency, but we currently strip the alpha channel from the output.

AI usage disclosure: YES.

Depends on https://github.com/Acly/comfyui-tooling-nodes/pull/71


Danamir:
Testing it. It seems to work.

Found a small optimization : in Edit mode, set `Text Encode Qwen Image 2.1` resolution to `0` do disable the internal resampling. Otherwise you can have a small pixel drift when editing. 

Note : even with 0 there is still an internal resampling to multiple of 32, so a custom selection can still be resized and provoke pixel drift. I don't think there is a way to entirely disable the resampling.

Danamir:
I tested by altering the `TextEncodeQwenImage21` code directly in ComfyUI. In `nodes_qwen.py` `execute` method, replacing :

```python
            if resolution > 0:
                ratio = samples.shape[3] / samples.shape[2]
                width = round(math.sqrt(resolution * resolution * ratio) / 32) * 32
                height = round(math.sqrt(resolution * resolution / ratio) / 32) * 32
            else:
                width, height = round(samples.shape[3] / 32) * 32, round(samples.shape[2] / 32) * 32
```
by 
```python
            if resolution > 0:
                ratio = samples.shape[3] / samples.shape[2]
                width = round(math.sqrt(resolution * resolution * ratio) / 32) * 32
                height = round(math.sqrt(resolution * resolution / ratio) / 32) * 32
            else:
                # width, height = round(samples.shape[3] / 32) * 32, round(samples.shape[2] / 32) * 32
                width, height = samples.shape[3], samples.shape[2]  # Don't resample if resolution == 0
```

Completely solves the pixel drift when using 0 as resolution. I suppose we would have to provide a custom node if we want this in the plugin.

Danamir:
The Fill method is not working if I already have content in my selection. I have to manually delete the pixels, or fill the selection with green, for the fill to generate truly new content.

Danamir:
Also worth noting : adding a `ModelSamplingAuraFlow` node with `shift = 6.0` on Qwen 2.1 can dramatically improve the rendering.

Danamir:
There is a bug in Refine mode : you have to disconnect the image from the reference input of `Text Encode Qwen Image 2.1`, otherwise it switches to reference mode instead of text to image mode, and the reference is overpowering the output. 

To test, do a 50% strength refine. It outputs the source image verbatim. If you disconnect the reference you can do refine normally.

For example in `workflow.py` `refine` method, set `image` parameter to `None` : 

```python
  prompt = encode_prompt(w, cond, clip, regions, None, vae, sampling=sampling)
```

Same thing for `refine_region`.

[Edit] : Wait, that breaks the Edit mode entirely. You have to remove the reference **only** when you are in Refine mode.

Danamir:
Your last commit broke the Edit, like I did with my too hasty fix. The reference is missing in Edit mode now.

fukc-gihtub:
> Also worth noting : adding a `ModelSamplingAuraFlow` node with `shift = 6.0` on Qwen 2.1 can dramatically improve the rendering.

[This page](https://comfy.org/workflows/image_qwen_image_2512_with_2steps_lora-6332ca621968/) says "The ModelSamplingAuraFlow node is crucial for sampling and refining the image output, ensuring that the final result aligns with the user's prompt."
Still, krita-ai-diffusion doesn't use that node for Qwen Image 2512.

Acly:
NB probably going to make a separate PR it will be easier.

Acly:
moving to #2617 - sorry to take over like that, but I'm short on time atm and it was faster to just redo it

fukc-gihtub:
np

